PR #1, opened 2015-05-20
OSS push from bbpcode repository

tribal-tec:


CMake/.gitignore
```diff
@@ -0,0 +1 @@
+Makefile
```
eile: ?

tribal-tec: ? The one from Buildyard. That's not checked in imo

eile: Me confused. BY does not put a Makefile in CMake/.

eile:
+1

CMake/.gitignore
tribal-tec: Ah, CMake/.gitignore. Dunno where this comes from, I remove it...

PR #7, opened 2015-07-06
Add '--format' for morphologyConverter app

mgeplf:
- Allows for choice of output format.  No change to the
  default at the moment (still h5v2), but allows for writing
  h5v1 files


eile:
@mgeplf: We will merge this in about a week, since we need to build the 2015.R2 module right now.

mgeplf:
Cool, no problem.  Let me know if the flag name isn't good, or any other changes that are needed.

apps/morphologyConverter.cpp
```diff
@@ -75,11 +77,19 @@ int main( int argc, char* argv[] )
         return EXIT_FAILURE;
     }
 
+    brion::MorphologyVersion out_version = brion::MORPHOLOGY_VERSION_UNDEFINED;
+    if ( vm["format"].as< std::string >() == "h5v1" ) {
+        out_version = brion::MORPHOLOGY_VERSION_H5_1;
+    }
```
eile: Style: no braces for one-liners, '{' on new-line for multi-liners.

eile: style: CamelCapitalization. But I would just name the variable 'format'

apps/morphologyConverter.cpp
```diff
@@ -115,7 +124,10 @@ int main( int argc, char* argv[] )
     readTime += clock.resetTimef();
 
     out.writeSectionTypes( *types );
-    out.writeApicals( *apicals );
+    if ( out.getVersion() != brion::MORPHOLOGY_VERSION_H5_1 ) {
+        out.writeApicals( *apicals );
+    }
```
eile: as above.

eile:
+1 otherwise

Issue #9, opened 2015-07-08
Brion morphology section types not fully compatible with SWC "spec"


juanchopanza:
The Brion morphology types defined in brion/enums.t `brion::enums::SectionType` do not match the ones defined in the SWC "specification", e.g. http://research.mssm.edu/cnic/swc.html:

```
    0 = undefined
    1 = soma
    2 = axon
    3 = dendrite
    4 = apical dendrite
    5 = fork point
    6 = end point
    7 = custom
```

Since the types are checked when parsing SWC files, valid files fail to parse. See for example `test_data/swc/sequential_trunk_off_1_16pt.swc` in [hbp-neurom's test data directory](https://bbpcode.epfl.ch/code/gitweb?p=algorithms/hbp-neurom.git;a=tree;f=test_data/swc;h=85e1e786fd29551140487736e948045c0058b147;hb=HEAD) (access to bbpcode.epfl.ch required.) In fact, all the `test_data/swc/sequential_trunk_off*.swc` are valid SWC.

hernando:
The enum in Brion is missing the types fork point, end point and custom. It's hard to decide what to do with this points and I don't understand why they were added to the specification in the first place, specially the end point (why these widespread standard file formats in Biology are so badly designed?). End points and fork points are implicit to the structure, making it optional to make them explicit only leaves room for consistency errors in my opinion.
I see that for branch points it might not be fully clear to which type they belong and hence the additional of a fork point, but for end points there should be no discussion. Then, what's the difference between undefined and custom? Having only one custom identifier is sort of useless.

I could add the extra types to the enum, but then we have to decide to do with them when they appear. Most probably this addition may break code with uses switch statements.

Regarding the example file. It may be syntactically correct, but not semantically for a few reasons. The most obvious are:
- The end point with id 7 is not and end point
- The fork points at ids 6 and 14 are have only one children, so they are not fork points (unless the for point tag actually means "hey, here there is a forking, but I don't care about one of the branches".
- Some section type transitions are biologically impossible: axon->dendrite, x->soma.
I don't know what sanity checks are implemented in MorphM, but I decided several in Brion, and following the same principle, consistency rules for the new types will be broken by this file. Giving up the sanity checks it's calling for trouble. In my opinion, bad input files should be detected as soon as possible (I can't remember how many times I've been blamed for errors that could be backtracked to the data).
That means that it won't be safe to simply add the new section types.

hernando:
[...]

PR #13, opened 2015-07-24
Issue #9: Extended SWC parser to understand all section types from the spec

hernando:
The Brion section types have not been extended to include the missing types,
instead fork and end points are converted into the most reasonable type
according to the parent point.


brion/morphology.h
```diff
@@ -61,7 +61,8 @@ class Morphology : public boost::noncopyable
      *  of the parent section.
      *
      * @param stage the repair stage to take the section information from
-     * @return index and parent index of all sections of the morphology
+     * @return First point index and parent index of all sections of the
```
eile: 'parent indices'?

brion/plugin/morphologySWC.cpp
```diff
@@ -36,12 +36,25 @@ namespace
...
+const int SWC_UNDEFINED_PARENT = -1;
+
+enum SWCSectionType
+{
+    SWC_SECTION_UNDEFINED = 0,
+    SWC_SECTION_SOMA,
+    SWC_SECTION_AXON,
+    SWC_SECTION_DENDRITE,
+    SWC_SECTION_APICAL_DENDRITE,
+    SWC_SECTION_FORK_POINT = 5,
```
eile: Why hard-coded 5?

hernando: I wanted to make it easy to identify which number corresponds to each tag. Instead of writing down all the value I just wrote that value (and now I'm writing longer to give explanations).

eile:
+1

eile:
What's up here?

hernando:
I was waiting for @juanchopanza to make any comment, but maybe he didn't get any notification from this PR

juanchopanza:
If I git it, I didn't see it. I can look at this in 2hrs or so.

juanchopanza:
+1

PR #19, opened 2015-08-27
Properly declare leaking Boost

tribal-tec:


tribal-tec:
FYI @hernando

rdumusc:
But is boost leaking? I see: list(APPEND BRION_DEPENDENT_LIBRARIES Lunchbox vmmlib)

tribal-tec:
Yes it is. Good catch! I wanted to doublecheck if this variable was set correctly.

rdumusc:
+1

PR #22, opened 2015-10-08
spikeConverter can process spikes to and from stream-type SpikeReport plugins

rdumusc:


apps/spikeConverter.cpp
```diff
@@ -29,11 +43,46 @@ int main( int argc, char* argv[] )
     }
 
     lunchbox::Clock clock;
```
bilgili: What about "--help" ? Better to do with program::options ?

apps/spikeConverter.cpp
```diff
@@ -29,11 +43,46 @@ int main( int argc, char* argv[] )
...
+        lunchbox::sleep( STREAM_SEND_DELAY_MS );
+
+        brion::Spikes outSpikes;
+        unsigned int frameIndex = 0;
+        for( auto spike : in.getSpikes( ))
+        {
+            if( spike.first > (frameIndex+1) * STREAM_FRAME_LENGTH_MS )
+            {
+                out.writeSpikes( outSpikes );
+                printSentFrame( frameIndex++, outSpikes.size( ));
```
bilgili: prefix ++ we do in general but no strong opposition. The assembly code may look similar in release mode. Otherwise it is a push stack and increase operation.

apps/spikeConverter.cpp
```diff
@@ -29,11 +43,46 @@ int main( int argc, char* argv[] )
...
+    brion::SpikeReport in( brion::URI( argv[1] ), brion::MODE_READ );
     const float readTime = clock.resetTimef();
 
     brion::SpikeReport out( brion::URI( argv[2] ), brion::MODE_WRITE );
-    out.writeSpikes( in.getSpikes( ));
+
+    // Stream-to-File conversion
+    if( in.getReadMode() == brion::SpikeReport::STREAM )
+    {
+        while( in.getNextSpikeTime() != brion::UNDEFINED_TIMESTAMP )
```
bilgili: you can merge in the while with an "&&"

apps/spikeConverter.cpp
rdumusc: It is intended as a postfix operation

bilgili: I am stupido :)

apps/spikeConverter.cpp
rdumusc: no, but I moved comments inside the if conditions as you requested

apps/spikeConverter.cpp
rdumusc: I can, but it was not the goal of this PR...

apps/spikeConverter.cpp
```diff
@@ -29,11 +43,49 @@ int main( int argc, char* argv[] )
...
+    else if( out.getReadMode() == brion::SpikeReport::STREAM )
+    {
+        // File-to-Stream conversion
+
+        // leave time for a Zeq connection to establish
+        lunchbox::sleep( STREAM_SEND_DELAY_MS );
+
+        brion::Spikes outSpikes;
+        unsigned int frameIndex = 0;
+        for( auto spike : in.getSpikes( ))
```
eile: afair Brion should still be C++03?

apps/spikeConverter.cpp
bilgili: You touch it, you own it ! ;)

rdumusc: Do I own Livre as well then..? ;-)

apps/spikeConverter.cpp
rdumusc: OK, will do!

apps/spikeConverter.cpp
bilgili: You guys owned it already ;) ( remember HBP )

apps/spikeConverter.cpp
```diff
@@ -29,11 +43,52 @@ int main( int argc, char* argv[] )
...
+            outSpikes.insert( *spikeIt );
+        }
+        out.writeSpikes( outSpikes );
+        printSentFrame( frameIndex, outSpikes.size( ));
+    }
+    else
+    {
+        // File-to-File conversion
+        out.writeSpikes( in.getSpikes( ));
+    }
```
eile: Not in this PR, (and I'm repeating myself), but the above code should not contain any 'isStream', but simply apply delays given on the command line to reading and writing spikes from/to the given source.

apps/spikeConverter.cpp
```diff
@@ -29,11 +43,52 @@ int main( int argc, char* argv[] )
...
-    out.writeSpikes( in.getSpikes( ));
+
+    if( in.getReadMode() == brion::SpikeReport::STREAM )
+    {
+        // Stream-to-File conversion
+        while( in.getNextSpikeTime() != brion::UNDEFINED_TIMESTAMP )
+        {
+            in.waitUntil( brion::UNDEFINED_TIMESTAMP, STREAM_READ_TIMEOUT_MS );
+            out.writeSpikes( in.getSpikes( ));
+        }
```
eile: Semantically the above is saying:
```
while next spike is defined [not undefinded]
    wait for next undefined spike
    write it
```
O_O

doc/Changelog.md
```diff
@@ -12,6 +12,9 @@ Changelog {#Changelog}
   Add brion::Target::parse() to resolve a given target name
 * [#12](https://github.com/BlueBrain/Brion/pull/12):
   Extended brion::Synapse to also support non-merged synapse files
+* [#22](https://github.com/BlueBrain/Brion/pull/22):
+  SpikeReport DSO plugins in the LD_LIBRARY_PATH are loaded automatically.
+  spikeConverter can process spikes to and from stream-type SpikeReport plugins.
```
eile: Make that two bullet items pointing to the same PR

brion/spikeReport.cpp
```diff
@@ -22,15 +22,20 @@
...
 
 #include <lunchbox/plugin.h>
 #include <lunchbox/pluginFactory.h>
 #include <lunchbox/uint128_t.h>
 #include <boost/scoped_ptr.hpp>
 
-namespace brion
+namespace
 {
+const std::string spikePluginDSONamePattern( "Brion.*SpikeReport" );
```
eile: Document this in class header of SpikeReport

apps/spikeConverter.cpp
hernando: in.waitUntil( brion::UNDEFINED_TIMESTAMP ) is the way to say wait until the end of the stream.
in.getNextSpikeTime returns brion::UNDEFINED_TIMESTAMP when the end has been reached.
So what you should be reading is

   while not finished
       wait until end with timeout
       write current.

The mistake I see is that writeSpikes is going to repeat all previously read spikes every iteration (as in 1 2 1 2 3 4 1 2 3 4 5 6...).  To way to avoid it is to call clear after each writeSpikes call.

A more clear way of doing the same would be providing new functions, but it isn't worth doing it if we plan to change the whole API to replace the std::multimap with a custom containers with magic iterators or something really low level. I didn't add those functions from the beginning because they are sort of redundant and preferred to keep the API lean.
I'm reluctant to write a low level API and prefer the iterator based one because I think that we are not going to get the low level API right for the stream and file based cases combined.

rdumusc: Correct, the spikes were being written multiple times and a clear() solved the problem, thanks!

doc/Changelog.md
rdumusc: done

brion/spikeReport.cpp
rdumusc: done

apps/spikeConverter.cpp
rdumusc: I don't think removing the isStream() is feasible given the current API. Passing the delay as command line could be done, but right now the fixed value is enough for our simple testing need IMHO...

eile: Agreed on both points, since isStream should be removed from the API, at least for what it is used now.

brion/spikeReport.h
```diff
@@ -81,6 +81,8 @@ class SpikeReport : public boost::noncopyable
      *        - NEST ('gdf' extension). NEST file based reports. In read mode,
      *          shell wildcards are accepted at the file path leaf to load
      *          multiple report files.
+     *        Support for additional types can be added thourgh plugins; see
```
eile: typo through

eile:
+1 from me

brion/spikeReport.h
rdumusc: done

PR #25, opened 2015-10-22
Add SpikeReport::getURI test

jafyvilla:
Remove doxygen in non-public headers
Remove redundant class members


tribal-tec:
the @version reported by @hernando is still missing?

jafyvilla:
As discussed with him, I removed all the doxygen from those headers, as they are not public and the copydoc doesn't add much information (it needs to be checked in the base class anyway). The compartment/morphology plugins don't have them either, so now we are consistent.

eile:
+1

PR #29, opened 2015-11-19
Additions required by Fivox

hernando:


README.md
```diff
@@ -37,6 +38,37 @@ Brion provides the following major features:
   [vmmlib](http://vmml.github.io/vmmlib),
   [Lunchbox](http://eyescale.github.io/Lunchbox-1.8/index.html),
   [Boost](http://www.boost.org/doc/libs).
+* Higher level classes:
+  * Read only morphology (brion::v2::Morphology)
+
+# Library versions {#Versions}
+
+Brion is evolving into a version 2 of its API. This new version will provide
+higher level classes to facilite queries and manipulation of the data model
```
bilgili: faciliate ?

brion/circuit.cpp
```diff
@@ -43,11 +43,12 @@ enum Section
     SECTION_CIRCUITSEEDS,
     SECTION_UNKNOWN
 };
+}
 
-class Circuit
+class Circuit::_Impl
```
bilgili: Do we need a _ before the class ?

brion/types.h
```diff
@@ -40,7 +40,7 @@
 #  include <vmmlib/vmmlib.hpp>
 #pragma warning(pop)
 
-/** @namespace brion Blue Brain File IO classes */
+/** @namespace brion Blue Brain File data model classes, types and functions. */
```
bilgili: BlueBrain or Blue Brain ? 99% of the time I am seeing BlueBrain :) not an issue though

brion/types.h
```diff
@@ -50,20 +50,52 @@ class Circuit;
...
+
+#ifdef BRION_USE_V2_API
+using namespace v2;
+#else
+using namespace io;
+#endif
+
 typedef vmml::vector< 2, int32_t > Vector2i; //!< A two-component int32 vector
 typedef vmml::vector< 3, float > Vector3f;   //!< A three-component float vector
 typedef vmml::vector< 4, float > Vector4f;   //!< A four-component float vector
```
bilgili: I think those shold b defined in vmml at a point. Everybody have them :)

brion/v2/morphology.cpp
```diff
@@ -0,0 +1,244 @@
...
+    SectionTypesPtr _types;
+    Vector2isPtr _apicals;
+
+    bool _isGlobal;
+    Vector3f _somaPosition;
+
+    _Impl( const boost::shared_ptr< const brion::Morphology >& morphology,
+           const MorphologyRepairStage stage )
+        : _morphology( morphology )
+        , _points( morphology->readPoints( stage ))
```
bilgili: do we have _ before variables or not ? IN the BlueConfig.cpp there was not.

rdumusc: Yes, _  before private members. In BlueConfig.cpp they are public, everything is correct

brion/blueConfig.cpp
```diff
@@ -137,18 +147,69 @@ class BlueConfig
...
+        const brion::Strings& runs = names[ brion::CONFIGSECTION_RUN ];
+        return runs.empty() ? std::string() : runs.front();
+    }
+
+    const std::string& get( const BlueConfigSection section,
+                            const std::string& sectionName,
+                            const std::string& key  ) const
+    {
+        // This function doesn't create entries in the tables in case they
+        // doesn't exist.
```
rdumusc: don't

brion/circuit.cpp
hernando: It's a private member, why not?

brion/types.h
hernando: All our copyright headers say "Blue Brain Project"

brion/v2/morphology.cpp
hernando: These members are also public. I kept the _ to make it clear they are members. But if you insist I can remove the _.

brion/types.h
eile: https://github.com/Eyescale/vmmlib/blob/master/vmmlib/vector.hpp#L408 

simply using vmml::Vector2i; should suffice. If some are missing in the above they should be added.

brion/v2/morphology.cpp
eile: _ is for private. Some Impls have private members.

README.md
```diff
@@ -37,6 +38,37 @@ Brion provides the following major features:
...
+In order to use Brion 2 there is nothing special to be done to compile Brion
+itself. In client code you have to define the macro \c BRION_USE_V2_API to
+enable this version. When this macro is set, all the new high level classes
+will live in the namespace \c brion and classes which clash with new ones are
+moved into the \c brion::io If the macro is not enabled, version 2 classes
+will still be available inside the namespace \c brion::v2.
+
+There is no need to recompile Brion to switch between versions. It is
+guaranteed that this macro only changes the namespace layout and that class
+interfaces remain equal.
```
eile: Just checking if I understood the motivation behind:
* Morphology will become io::Morphology in the V2 API
* v2::Morphology will become the new Morphology class
?

brion::io seems redundant, maybe we can find a better namespace for the v2 classes.

brion/blueConfig.h
```diff
@@ -92,6 +92,32 @@ class BlueConfig : public boost::noncopyable
     }
     //@}
 
+    /** @name Semantic read API */
+    //@{
+    /** @return the full path to the circuit.mvd2 file. @sa Circuit */
```
eile: in the light of mvd3 (and potentially ZeroEQ-accessible circuits), shouldn't this be 'the URI to the circuit information'?

README.md
hernando: Right and right.
You mean that io is redundant right now or in the future? If io is going to be in v2 it must be also now so you can already write v2 code.

Sure, this is my proposal. If you have a better one, your welcome. The main question is the following: which are the first class citizen classes in v2, the low level ones or the new ones.

brion/circuit.cpp
eile: It's a type :P (just bean counting, not a reply to your question)

brion/io/morphologies.cpp
```diff
@@ -0,0 +1,136 @@
...
+{
+    const brion::Circuit circuit( config.getCircuitSource( ));
+    return load( circuit, config.getMorphologySource(), gids );
+}
+
+Morphologies load( const BlueConfig& config, GIDSet& gids )
+{
+    brion::Circuit circuit( config.getCircuitSource( ));
+    return load( circuit, config.getMorphologySource(), gids );
+}
```
eile: const-only overloads on references are dangerous to user-level code, as the non-const is afaik preferred.

README.md
eile: Redundant: brion::io is redundant due to Br_io_n

First class citizen: I feel that the current brion API (minus SpikeReport ;) ) is at the right level for an IO lib. The new stuff probably needs a new name, even project/top-level namespace. To be discussed.

brion/io/morphologies.cpp
hernando: We've had so many times the discussion about throw versus warn that I decided to implement the two flavours. The non-const version being preferred is consistent with not throwing by default, so unless you've changed your mind about that I'll keep it like this. It also puts the burden of verifying that the set sizes are the same on the client, but that's the price you have to pay for not throwing. It's a matter of trade-offs and I don't know if there's a one size fits all.

eile: I simply raised it since I did the same for the Publisher( uri ) ctor, which @tribal-tec reverted for good reasons.
There are alternatives, mostly by using different function signatures (return a std::pair, provide an error function, ...)

brion/circuit.cpp
hernando: OK, being pedantic it's a nested types, as it seems that's the term used in the standard instead of "member type".

brion/io/morphologies.cpp
hernando: What I want these functions to provide is a simple way to load a set of morphologies from a GIDSet where:
- It's easy to relate each GID to a morphology.
- If it fails it's possible to know for which GID/s it failed
- May or may not throw.
Simple (apparently).

Actually the error functor is a good idea as it satisfies all three requirements. I can think of the following specification for the functor:
- Takes a GID as input
- May throw or not
- The return type may be void, in which case a null pointer is pushed on the return value.
- The return type may be bool, in which case the load function will push a null pointer only if the error functor returns true.

One shortcoming of this approach is that the functor parameter type has to be either a template or a virtual class. I don't like the  idea of using a virtual class, and using a template forces to move the implementation to a public header. A hybrid solution is to have a template function that instantiates a template class that inherits from the virtual functor class, this way you can hide the implementation and still have the template, but I find this solution a bit of overdesign.

Another aspect is that I still want to make it easy to relate the loaded morphologies to the GIDs in the non-exceptional case, and the best way to do it without returning an additional GIDSet is to ensure that the iteration order of the morphology list follows the iteration order of the intput GIDSet. Then, if a morphology failed to load we have to include an null morphology in the list and that can only be done if Morphologies is a list of pointers. Indeed, right now it's a list of pointer, but just to be aware that it won't be possible to change it later without having to change the whole thing.

README.md
tribal-tec: I have the same feeling. I was hoping we would have such discussions not again at the pull request level. If I look at it twice, it's just another SDK-like API. It's either an auxiliary library in Brion or even a separate project. This V1/V2 define style is not the right approach imo.

brion/blueConfig.cpp
```diff
@@ -65,6 +68,12 @@ inline std::string lexical_cast( const brion::BlueConfigSection& b )
 }
 }
 
+namespace
+{
+    const char* const SPIKE_FILE = "/out.dat";
```
tribal-tec: shall we group those semantics in a common place and document them once and for all so that everybody knows that they exist? Same for the strings used below like "CircuitTarget" etc.

brion/blueConfig.cpp
```diff
@@ -137,18 +147,69 @@ class BlueConfig
...
+        const KVStore& store = tableIt->second;
+        const KVStore::const_iterator kv = store.find( key );
+        if( kv == store.end( ))
+            return empty;
+        return kv->second;
+    }
+
+    const std::string& getCircuitTarget()
+    {
+        return get( brion::CONFIGSECTION_RUN, getRun(), "CircuitTarget" );
```
tribal-tec: On one hand, we use a define for "Run" section, but for an explicit entry it's a magic string??

brion/blueConfig.cpp
```diff
@@ -162,17 +223,96 @@ BlueConfig::~BlueConfig()
...
                                     const std::string& sectionName,
                                     const std::string& key ) const
 {
-    return _impl->_table[section][sectionName][key];
+    return _impl->get( section, sectionName, key );
+}
+
+std::string BlueConfig::getCircuitSource() const
+{
+    return get( CONFIGSECTION_RUN, _impl->getRun(), "CircuitPath" ) +
```
tribal-tec: dito, group the semantics

brion/blueConfig.h
tribal-tec: That would be the case in general for all returned sources. spike and report source are URIs, but synapse and circuit sources are strings btw, should be the same for all.

brion/blueConfig.h
```diff
@@ -92,6 +92,32 @@ class BlueConfig : public boost::noncopyable
...
+    BRION_API brion::URI getReportSource( const std::string& report ) const;
+
+    /** @return the URI to the default spike data source. @sa SpikeReport */
+    BRION_API brion::URI getSpikeSource() const;
+
+    /** @return the set of GIDs for the given target, or the circuit target if
+     *          the given target is empty. */
+    BRION_API brion::GIDSet parseTarget( std::string target ) const;
+
+    /** @return the simulation timestep or NaN if undefined. */
```
tribal-tec: timestep in ms

brion/circuit.cpp
tribal-tec: First time we do _Impl, I also don't see the point to deviate here. The Impl is by definition private, so _ would be correct, but it's also rather redundant. Change it everywhere else and I'm fine with it :)

brion/io/morphology.cpp
```diff
@@ -28,36 +28,34 @@
...
     }
 
     boost::scoped_ptr< MorphologyPlugin > plugin;
 };
-}
 
 Morphology::Morphology( const std::string& source )
-    : _impl( new detail::Morphology(
-                 MorphologyInitData( lunchbox::URI( source ))))
+    : _impl( new _Impl( MorphologyInitData( lunchbox::URI( source ))))
```
tribal-tec: while touching this code, replace lunchbox::URI with servus::URI

tribal-tec: while touching this code, replace lunchbox::URI with servus::URI

brion/io/plugin/morphologyHDF5.cpp
```diff
@@ -52,6 +52,8 @@ inline std::string lexical_cast( const brion::MorphologyRepairStage& s )
 
 namespace brion
 {
+namespace io
```
tribal-tec: the io namespace really does not make sense to me.

brion/v2/morphology.cpp
```diff
@@ -0,0 +1,244 @@
...
+        , _apicals( morphology->readApicals( ))
+        , _isGlobal( false )
+        , _somaPosition( Vector3f::ZERO ) // the soma is assumed to be
+                                          // centered at the 0, 0, 0
+    {
+    }
+
+    uint32_ts getSectionIDs( const SectionTypes& types ) const
+    {
+        bool bits[] = { false, false, false, false, false };
```
tribal-tec: std.bitset?

brion/v2/morphology.cpp
```diff
@@ -0,0 +1,244 @@
...
+        }
+
+        const size_t start = ( *_sections )[sectionID][0];
+        const size_t end = sectionID == _sections->size() - 1 ?
+            _points->size() : ( *_sections )[sectionID + 1][0];
+
+        Vector4fs result;
+
+        // Dealing with the degenerate case of single point sections
+        assert(end > start);
```
tribal-tec: in the light of broken morphologies, I would refrain of asserting things when they crash your program. Report clearly that there is a mistake with this morphology.

brion/v2/morphology.cpp
```diff
@@ -0,0 +1,244 @@
...
+            for( size_t i = 0; i != points.size(); ++i)
+                result.push_back(( *_points )[start] );
+
+        // Computing the accumulated length at each segment of the section
+        std::vector< float > accumLengths;
+        accumLengths.push_back(0);
+        for( size_t i = start; i != end - 1; ++i )
+            accumLengths.push_back(
+                accumLengths.back() +
+                (( *_points )[i + 1] - ( *_points )[i]).length( ));
```
tribal-tec: not really readable. scope and some variables?

doc/Changelog.md
```diff
@@ -1,8 +1,14 @@
 Changelog {#Changelog}
 =========
 
-# git master {#master)
+# git master
 
+* [29](https://github.com/BlueBrain/Brion/pull/29):
+  - First steps towards v2 of the API.
```
tribal-tec: '*' instead of '-' ?

tests/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@
 #
 # This file is part of Brion <https://github.com/BlueBrain/Brion>
 #
-# Change this number when adding tests to force a CMake run: 1
+# Change this number when adding tests to force a CMake run: 2
```
tribal-tec: not needed as you touched existing test files

brion/io/morphology.cpp
eile: Or have a using servus::URI; in types and just use URI?

eile:
> I was hoping we would have such discussions not again at the pull request level.

For me, the point of this PR was not to have these discussions, but to have a base upon which we can discuss. In which format should we do this? Skype tomorrow?

brion/circuit.cpp
hernando: Removed

brion/io/morphology.cpp
hernando: There was a using lunchbox::URI, I've changed it to servus::URI

README.md
hernando: Of course it's an SDK-like API, that's completely unavoidable because that's what it's trying to replace.
Since we didn't discuss how to deal with this I decided to follow an incremental approach to it, and actually that's what I thought @eile and me have agreed. A new library it's a good idea and for the moment I'd keep it in this project. I just hope it doesn't block this PR because we can't agree on a name for it. The new library can host the future SpikeReport.
It's painful that these questions emerge and are decided in PRs, you can do better but we need to spend more time in planning or involve more people.

bilgili: I also did not like the V1/V2 seperation using macros.. May not be another way ? Instead of macros, detecting the situation and instantiating the right implementation wthout the user noticing v1/v2 ? Or, it is complicated for this pr ?

hernando:
I think that it's clear that a separate library is preferred by everybody. For the moment I suggest to name it libBrain (Brion's algorithmic interface), we can rename it before the final merge.

brion/blueConfig.cpp
hernando: I'm working on this

brion/io/morphologies.cpp
hernando: You haven't commented on this answer.

brion/morphologies.h
```diff
@@ -0,0 +1,109 @@
...
+
+
+/**
+ * Loads a set of morphologies.
+ *
+ * Throws if any of the morphologies could not be loaded.
+ * @param paths The path for the morphologies to load.
+ * @return The list of morphologies in the iteration order of paths.
+ */
+Morphologies load( const StringSet& paths );
```
hernando: Should the input be a URISet for consistency with other methods?

brion/blueConfig.cpp
hernando: I moved all the definitions here but I guess you're referring to another place out of this file, possibly a public header, right?

tribal-tec: Public header sounds good. People and tests can use it directly as well then w/o the convenience functions if they want to.

brion/types.h
```diff
@@ -53,16 +53,19 @@ class Mesh;
...
 class SynapseSummary;
 class Target;
 
-typedef vmml::vector< 2, int32_t > Vector2i; //!< A two-component int32 vector
-typedef vmml::vector< 3, float > Vector3f;   //!< A three-component float vector
-typedef vmml::vector< 4, float > Vector4f;   //!< A four-component float vector
-typedef vmml::vector< 3, double > Vector3d; //!< A three-component double vector
-typedef vmml::vector< 4, double > Vector4d; //!< A four-component double vector
+typedef boost::shared_ptr< Morphology > MorphologyPtr;
+typedef std::vector< MorphologyPtr > Morphologies;
```
tribal-tec: used in brain, not brain, right?

brion/types.h
```diff
@@ -124,7 +127,9 @@ const float RESTING_VOLTAGE = -67.; //!< Resting voltage in mV
 const float MINIMUM_VOLTAGE = -80.; //!< Lowest voltage after hyperpolarisation
 
 using lunchbox::Strings;
-using lunchbox::URI;
+typedef std::set< std::string > StringSet;
```
tribal-tec: dito

brion/types.h
hernando: I kept the morphologies::load functions in brion. I think they still belong there.

brain/morphology.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+namespace
+{
+    void nullDeleter( const brion::Morphology* ) {}
+}
+
+class Morphology::Impl
+{
+public:
+    boost::shared_ptr< const brion::Morphology > _morphology;
```
tribal-tec: shared with whom?

brion/types.h
tribal-tec: Ok, I thought it was loading brain morphologies. This does not exist?

brain/morphology.cpp
hernando: This is mainly motivated by the prototype of brion::morphologies::load (which is still RFC). This set of functions returns std::vector< boost::shared_ptr<brion::Morphology> > because brion::Morphology is non copyable and we can't use movables from C++11 (in case CoreNeuron finally adopts it, which implies Brion has to compile in BG).

With this shared_ptr you can have the 3 possibilities provided by the brain::Morphology constructors:
* Create a brion::Morphology right away
* Take a shared pointer
* Take a reference (which is stored as a shallow shared_ptr with null deallocator)
I could also make brain::Morphology always take a const brion::Morphology* to imply that the constructor will store a reference, but that puts the whole responsibility of managing the memory on the caller and I find it less convenient.

brion/types.h
hernando: Now that I think about it. It also makes sense to have a brain::morphologies::load function to load morphologies that can be optionally translated into global coordinates automatically. This function will be more efficient and usable that what I did in Fivox.

brain/morphology.h
```diff
@@ -0,0 +1,122 @@
...
+#include <brain/api.h>
+#include <brain/types.h>
+
+#include <boost/noncopyable.hpp>
+
+namespace brain
+{
+
+/**
+ *  A higher level interface to morphological data than the raw morphology.
```
tribal-tec: A more elaborate description is needed here. With this statement I'm not convinced to use this API over the other one.

brain/morphology.h
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * @version usntable
```
tribal-tec: typo usntable

brain/morphology.h
```diff
@@ -0,0 +1,122 @@
...
+     * Create the morphology and load all the data.
+     */
+    BRAIN_API explicit Morphology(
+        const std::string& source,
+        brion::MorphologyRepairStage stage = brion::MORPHOLOGY_UNDEFINED );
+    /**
+     * Create the morphology and load all the data.
+     */
+    BRAIN_API explicit Morphology(
+        const brion::Morphology& morphology,
```
tribal-tec: For simplicity in the API and the implementation, is this really needed? It reads so weird that a morphology takes a morphology for construction. Kind of a copy constructor, but not really. I would vote to keep the string (why no URI btw?) ctor only.

brain/morphology.h
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Create the morphology and load all the data.
+     */
+    BRAIN_API explicit Morphology(
+        const brion::MorphologyPtr& morphology,
+        brion::MorphologyRepairStage stage = brion::MORPHOLOGY_UNDEFINED );
+
+    BRAIN_API ~Morphology();
+
+    BRAIN_API const brion::Vector4fs& getPoints() const;
```
tribal-tec: doxygen?

brain/morphology.cpp
```diff
@@ -0,0 +1,263 @@
...
+            accumLengths.push_back( accumLengths.back() +
+                                    ( segmentEnd - segmentStart ).length( ));
+        }
+        const float totalLength = accumLengths.back();
+
+        for( brion::floats::const_iterator i = points.begin();
+             i != points.end(); ++i )
+        {
+            // Finding the segment the requested sampling position falls in
+            float length = std::max( 0.f, std::min( 1.f, *i )) * totalLength;
```
tribal-tec: const

brain/morphology.cpp
```diff
@@ -0,0 +1,263 @@
...
+            size_t index = 0;
+            while( accumLengths[index + 1] < length &&
+                   index < accumLengths.size() - 1 )
+                ++index;
```
tribal-tec: scope for readability?

brain/morphology.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+            // Interpolating the point
+            const float alpha = ( length - accumLengths[index] ) /
+                              ( accumLengths[index + 1] - accumLengths[index] );
+            const brion::Vector4f sample =
+                ( *_points )[start + index + 1] * alpha +
+                ( *_points )[start + index] * (1 - alpha );
+            result.push_back( sample );
```
tribal-tec: it's possible to reserve() the result? Or the amount of points is always small enough?

brain/morphology.cpp
```diff
@@ -0,0 +1,263 @@
...
+brion::Vector4fs Morphology::getSectionSamples(
+    const size_t sectionID, const brion::floats& points ) const
+{
+    return _impl->getSectionSamples( sectionID, points );
+}
+
+void Morphology::transform( const brion::Circuit& circuit, const uint32_t gid )
+{
+    if( _impl->_isGlobal )
+        throw std::logic_error(
```
tribal-tec: LBTHROW

brain/morphology.cpp
```diff
@@ -0,0 +1,263 @@
...
+                         boost::lexical_cast< float >( data[0][2] )));
+
+    transform( matrix );
+    _impl->_isGlobal = true;
+}
+
+void Morphology::transform( const brion::Matrix4f& matrix )
+{
+    if( _impl->_isGlobal )
+        throw std::logic_error(
```
tribal-tec: dito

brion/morphologies.h
```diff
@@ -0,0 +1,115 @@
...
+
+#include <brion/api.h>
+#include <brion/types.h>
+
+namespace brion
+{
+
+/** @namespace brion::morphologies
+ *  This namespace contains several free functions for loading sets of
+ *  morphologies.
```
tribal-tec: This convenience of loading is almost the same we have with the SDK today. Together with its shortcomings. In-core and no lazy-loading. Do we really gain much by providing those? I agree on having some convenience to construct the paths/URIs for loading from a folder for instance, but the actual loading should be done by the application/user-code right now.

brain/morphology.h
hernando: I have a reason for keeping at least the constructor that takes a MorphologyPtr, which is to be able to reuse the output of brion::morphologies::load.  I'm not using the constructor that takes a const brion::Morphology&, so it could be removed. I added simply for convenience.

tribal-tec: You should then use brian::morphologies::load then. Not a valid reason to me still. Convince me again :)

brain/morphology.cpp
hernando: Can be reserved.

brain/morphologies.h
```diff
@@ -0,0 +1,96 @@
...
+ *        are removed from the set.
+ * @return The list of loaded morphologies in the iteration order of the
+ *         output gids and transformed to circuit global coordinates.
+ * @version 1.7
+ */
+BRAIN_API Morphologies load( const brion::BlueConfig& config,
+                             const brion::GIDSet& gids );
+
+/**
+ * Convenience overload of the function above that takes an immutable
```
rdumusc: the doc of the two functions is inverted, this one has the mutable set

brain/morphology.h
hernando: brain::morphologies::load uses this constructor in the implementation, I'd need to either refactor and expose the implemenation or repeat it here if I don't have this constructor.

(BTW: brain or brian? Which one do you prefer?)

brain/morphologies.cpp
```diff
@@ -0,0 +1,100 @@
...
+        gids.size(), brion::Matrix4f::IDENTITY );
+    #pragma omp parallel for
+    for( size_t i = 0; i < gids.size(); ++i )
+    {
+        brion::Matrix4f& matrix = transforms[i];
+        matrix.rotate(
+            boost::lexical_cast< float >( data[i][3] ) * ( M_PI/180.0f ),
+            brion::Vector3f( 0, 1, 0 ));
+        matrix.set_translation(
+            brion::Vector3f( boost::lexical_cast< float >( data[i][0] ),
```
bilgili: catching the throw on lexcast needed ?

brain/morphologies.cpp
```diff
@@ -0,0 +1,100 @@
...
+        morphology->transform( transforms[i] );
+        morphology->setGlobal();
+        result.push_back( morphology );
+    }
+    return result;
+}
+
+}
+
+Morphologies load( const brion::BlueConfig& config, const brion::GIDSet& gids )
```
bilgili: this has been discussed I know but !on the API level having two function names similar doing different things is weird. It can be loadToGIDs and loadFromGIDs instead of being confusing ?

brain/morphologies.h
```diff
@@ -0,0 +1,96 @@
...
+ *  This namespace contains several free functions for loading sets of
+ *  morphologies in circuit global coordinates
+ */
+namespace morphologies
+{
+
+/**
+ * Load morphologies given a BlueConfig and GID set.
+ *
+ * @param config The object representing the BlueConfig or CircuitConfig with
```
rdumusc: not CircuitConfig in this function

brain/morphology.cpp
```diff
@@ -0,0 +1,263 @@
...
+            // Interpolating the point
+            const float alpha = ( length - accumLengths[index] ) /
+                              ( accumLengths[index + 1] - accumLengths[index] );
+            const brion::Vector4f sample =
+                ( *_points )[start + index + 1] * alpha +
+                ( *_points )[start + index] * (1 - alpha );
+            result.push_back( sample );
+        }
+
+        return result;
```
bilgili: return moved data ?
return std::move(result) and function signature is Vect4f&& ... ?

brain/morphology.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+    const brion::NeuronMatrix& data = circuit.get(
+        set, brion::NEURON_POSITION_X | brion::NEURON_POSITION_Y |
+             brion::NEURON_POSITION_Z | brion::NEURON_ROTATION );
+
+    brion::Matrix4f matrix( brion::Matrix4f::IDENTITY );
+    matrix.rotate( boost::lexical_cast< float >( data[0][3] ) * ( M_PI/180.0f ),
+                   brion::Vector3f( 0, 1, 0 ));
+    matrix.set_translation(
+        brion::Vector3f( boost::lexical_cast< float >( data[0][0] ),
```
bilgili: throw needed ?

brion/morphologies.h
hernando: I don't fully agree with your view. In the spirit of DRY I think these functions should be provided regardless of having others for out-of-core and lazy loading. Without them I'd have already repeated code in Fivox, which is not good.
On the other hand, I could make these functions return lists of paths, which solves the problem of how to report IO errors, and leave the convenience function for brain::morphologies. That's something I'd agree.

Actually this functions don't really load anything because all the accessors provided by brion::Morphology do lazy loading. The class that could do lazy loading but it's not is brain::Morphology.

brion/circuit.h
```diff
@@ -26,8 +26,6 @@
 namespace brion
 {
 
-namespace detail { class Circuit; }
-
 /** Read access to a Circuit file.
  *
  * Following RAII, this class is ready to use after the creation and will ensure
```
bilgili: this RAII keyword information is really needed I wonder. I never saw in another API doc, or thats only me.

brion/morphologies.cpp
```diff
@@ -0,0 +1,134 @@
...
+    {
+        const std::string path =
+            database.getPath() + "/" + matrix[index][0] + ".h5";
+        boost::shared_ptr< Morphology >& morphology = loaded[path];
+        if( !morphology )
+            morphology.reset( new Morphology( path ));
+
+        result.push_back( morphology );
+    }
+    return result;
```
bilgili: ditto move, but if it is C++11 ?

brain/morphologies.h
hernando: I'm referring to the BlueConfig or CircuitConfig files. Clarified.

brain/morphologies.h
hernando: done

brain/morphologies.cpp
hernando: Good point! The problem is that I cannot throw from an OMP loop, so I will simply print a warning a reset the transform to the identity.

tests/data/h5/createNeuron.py
```diff
@@ -0,0 +1,67 @@
...
+SOMA = 1
+AXON = 2
+BASAL = 3
+APICAL = 4
+
+out = h5py.File('test_neuron.h5', 'w')
+
+def createTBranchPoints():
+    # Creates a branch with a T shape
+    # The first section is a non-homogeneous sampling of 10 points on a straight
```
bilgili: distances between lines are non-homogeneous, the points themselves or the sampling behaviour are homogeneous so it may be differently said ? Non-homogenus distances based on sampling. I am not totally sure you may be also right but it kind of itched my ear :)

tests/morphologies.cpp
```diff
@@ -0,0 +1,210 @@
...
+    for( uint32_t gid = 1; gid < 500; gid += 75)
+        gids.insert(gid);
+    const size_t size = gids.size();
+    brion::Morphologies morphologies =
+        brion::morphologies::load( config, gids );
+    BOOST_CHECK_EQUAL( morphologies.size(), size );
+    BOOST_CHECK_EQUAL( gids.size(), size );
+
+    // Checking the first morphology
+    _checkMorphology( *morphologies[0], "R-C010306G.h5" );
```
bilgili: we should have an r2d2.h5 :)

brain/morphology.cpp
hernando: No C++11 for Brion for the moment.

brion/circuit.h
hernando: I found 5 coincidences in my Boost headers :P
I'll rephrase it.

tests/data/h5/createNeuron.py
hernando: They are not homogeneous on purpose. The sampling I do in the test uses homogeneous distances because it's easier to verify. If the segments here are also homogeneous I thought that the probability of passing the test by chance increases.

tests/morphologies.cpp
hernando: That could have been the name of the test morphology.

brion/morphologies.h
hernando: Do yo agree on having a Circuit method with this prototype:

     URIs getMorphologyURIs( const URI& database, const GIDSet& gids ) const

and then remove all brion::morphologies::load functions and only keep the brain::morphologies::load ones.

tests/data/h5/createNeuron.py
bilgili: Nope I am not against what you did. The use of phrase "non-homogeneous sampling of points" bothered me :) "Non-regular sampling" yes, "Non-homogeneous distance sampling" yes but sampling of point non-homogenus was a bit mis-leading.

hernando: "Non-homogeneous distance sampling" may be clearer

bilgili: Exactly

brain/morphology.h
eile: The life of Brian... any sufficiently advanced reality approaches satire?

brion/morphologies.h
rdumusc: I'm starting to get my head around this change :-) Your new proposition looks better. Especially since brion::Morphology are really MorphologyReader/Writer classes, they don't load any data until told to do so. If there would be a need for this brion::morphology::load() it should be named brion::morphology::open() it seems.

brain/morphologies.cpp
```diff
@@ -0,0 +1,114 @@
...
+                                 boost::lexical_cast< float >( data[i][1] ),
+                                 boost::lexical_cast< float >( data[i][2] )));
+        }
+        catch( boost::bad_lexical_cast& e)
+        {
+            LBWARN << "Error parsing circuit position or orientation for gid "
+                   << *gid << ". Morphology not transformed." << std::endl;
+            matrix = brion::Matrix4f::IDENTITY;
+        }
+        #pragma omp critical (transform_gid)
```
tribal-tec: iirc, needs to be unique across the entire application. brion_transform_gid?

brain/morphology.cpp
```diff
@@ -0,0 +1,278 @@
...
+    {
+        matrix.rotate(
+            boost::lexical_cast< float >( data[0][3] ) * ( M_PI/180.0f ),
+            brion::Vector3f( 0, 1, 0 ));
+        matrix.set_translation(
+            brion::Vector3f( boost::lexical_cast< float >( data[0][0] ),
+                             boost::lexical_cast< float >( data[0][1] ),
+                             boost::lexical_cast< float >( data[0][2] )));
+    }
+    catch( boost::bad_lexical_cast& )
```
tribal-tec: const ref

brain/morphologies.cpp
```diff
@@ -0,0 +1,114 @@
...
+        ++gid;
+    }
+
+    Morphologies result;
+    for( size_t i = 0; i < raw.size(); ++i )
+    {
+        MorphologyPtr morphology( new Morphology( raw[i] ));
+        morphology->transform( transforms[i] );
+        morphology->setGlobal();
```
tribal-tec: transform() sets the global state already, right?

hernando:
As agreed I've taken back the last commit and replaced it with some final polishing so this PR can be merged and another one can be created for the new library.

tribal-tec:
+1

tribal-tec:
Hold on, Changelog?

hernando:
It's updated

tribal-tec:
@rdumusc any last comment? merge otherwise

rdumusc:
nope, good to merge! Just rename the PR title before doing so? Not really linked to Fivox, right?

hernando:
It is linked to Fivox

PR #31, opened 2015-11-27
Fix crash while reading more than `ulimit -Sn` (1024 default) NEST gdf files

tribal-tec:


bilgili:
"The read/write being in the same class", I believe we have discussed before. Current situation is a bit smelly but changing it needs lots of change as well now.

hernando:
As smelly as std::fstream, or file in Python or FILE* handlers in C? I think that's a design decision (Java on the other hand has separate readers and writers) and I think in general there isn't anything inherently bad about it.

bilgili:
If you check how std::fstream is based, it is based on two seperate streamers. They had to the design desicion based on when it is both read and write. In case you have one read or write operation you use ifstream and ofstream. We took the desicion that everything is readable and writable. I am not saying we did the wrong thing, I think it is a bit smelly.

PR #35, opened 2015-12-14
Added new brain::Section class and related methods in brain::Morphology.

hernando:
The new class provides access to morphological data on a per
section basis. From a section it's possible to query the parent and
children as well as information for its sample points. The new
class is very lightweight so it can be created on-the-fly by
brain::Morphology and doesn't need to be stored anywhere.


hernando:
Something I'd like to discuss is whether we should consider that the soma is also a Section or not.

brain/morphology.cpp
```diff
@@ -61,7 +78,7 @@ class Morphology::Impl
         for( size_t i = 0; i != types->size(); ++i )
```
bilgili: for( const SectionType type: *types ) ?

brain/morphology.cpp
```diff
@@ -70,11 +87,35 @@ class Morphology::Impl
...
+            length = _computeSectionLength( sectionID );
+        return length;
+    }
+
+    Vector4fs getSectionSamples( const uint32_t sectionID ) const
+    {
+        const SectionRange range = getSectionRange( sectionID );
+        Vector4fs result;
+        result.reserve( range.second - range.first );
+        for( size_t i = range.first; i != range.second; ++i )
```
bilgili: std::copy( begin, end, back_inserter  ) or result.insert( ... ) ?

brain/morphology.cpp
hernando: None is possible (C++11 even less, because we can't use it in Brion).
Note that line 85 needs the index.

brain/morphology.cpp
```diff
@@ -131,14 +150,56 @@ class Morphology::Impl
...
+            // For the other cases it doesn't matter to have concurrent updates
+            // because they will yield the same result (and it's probably
+            // cheaper to go ahead with the computation than to contend for a
+            // mutex).
+            distance = getSectionLength( parent ) + getDistanceToSoma( parent );
+        }
+        return distance;
+    }
+
+    floats getDistancesToSoma( const uint32_t sectionID ) const
```
bilgili: To very similar names to functions both same parameter is a bit confusing ? Object can be both singular or plural ?

brain/section.h
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Return the absolute distances to the soma for the sample positions of
+     * this section.
+     *
+     * @return A list of distances. For a section consisting
+     *         of n segments, this list will have n + 1 values. The section
+     *         length is equal to the difference between the first and last
+     *         values of the list.
+     */
+    BRAIN_API floats getDistancesToSoma() const;
```
bilgili: getSegmentDistances()

brain/morphology.cpp
hernando: Do you really think any of the below is better?:

    result.insert( result.end(),
                   points->begin() + range.first, points->begin() + range.second)

    std::copy( points->begin() + range.first, points->begin() + range.second,
               std::back_inserter( result ));

brain/morphology.cpp
hernando: I partially agree, but I would argue that the return types are different, so you will notice you're not using the right one at compile time sooner or later. And the same happens with many of our types, e.g float vs floats.

brain/section.h
hernando: getSamplesDistances()?

I'd prefer to avoid the word segment as it doesn't appear in any other place in the API.

hernando: Also it should be getSamplesDistancesToSoma() which is a bit long in my opinion. getXXXDistances is not sufficiently explicit.

brain/morphology.cpp
bilgili: As a matter of fact :) Yep !

brain/section.h
bilgili: Yep, getSubDistancesToSoma() .. or getSampleDistsToSoma() ?

eile:
+600 loc... ugh.

There are plenty of new functions in morphology.cpp which do not appear in morphology.h? I got lost in the naming and wanted to read doxygen first - unless I missed something, can you make this PR more understandable, or I will -1 it on grounds of fighting potential SDK dragons. ;)

hernando:
I can do as you request by moving the detail::Morphology into it's separate .h/.cpp files. Originally detail::Morphology was Morphology::Impl, but the implementation of Section needs to access the Morphology internals so I had to move it to a public class (I preferred that to making it friend). I kept evertyhing in one file being conscious that someone could complain, but there wasn't any example of this pattern I could use as a reference. If I do this change morphology.cpp will be split into 3 files.

On the other hand -1 without actual facts on what's wrong, only on the assumption that something might be wrong is not fair. That would be an authority veto, which is something I don't agree with.

eile:
> the implementation of Section needs to access the Morphology internals 

That's what I missed. Weird pattern. In that case, these methods should be considered public and documented. For example, I have no clue what getSectionSamples does.

eile:
-1 would be on grounds that I don't understand, not on authority (which is wrong). Apologies of putting you on the spot, but I want to finally understand this stuff and to avoid complexity where possible.

eile:
> the implementation of Section needs to access the Morphology internals

Not having understood the code: maybe the relevant methods should be part of the Morphology API?

brain/section.h
```diff
@@ -0,0 +1,135 @@
...
+ * A Section is an unbranched piece of a morphological skeleton.
+ * This class provides functions to query information about the sample points
+ * that conform the section and functions to obtain the parent and children
+ * sections.
+ *
+ * The cell soma is also considered a section, but some functions have
+ * special meaning for it.
+ *
+ * Sections cannot be directly created, but are returned by several
+ * brain::Morphological and brain::Section methods.
```
eile: brain::Morphology?

brain/section.h
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * This is a lightweight object with STL container style thread safety.
+ * It is also safe to use a section after the morphology from where it comes
+ * has been deallocated. The morphological data will be keept as long as there
```
eile: keept

brain/section.h
```diff
@@ -0,0 +1,135 @@
...
+
+    bool operator==( const Section& section ) const;
+
+    /** Return the ID of this section. */
+    uint32_t getID() const;
+
+    SectionType getType() const;
+
+    /**
+     * Return the total length of this section.
```
eile: in which unit?

brain/section.h
```diff
@@ -0,0 +1,135 @@
...
+     * If this section is a soma section the length is ill-defined and this
+     * function will return 0.
+     */
+    BRAIN_API float getLength() const;
+
+    /**
+     * Return the list of all point samples that define this section.
+     *
+     * If this sections is a soma section return the list of points in the
+     * soma ring.
```
eile: What is a soma ring? What can I do with this for a soma? Return an empty vector or just the center?

brain/section.h
```diff
@@ -0,0 +1,135 @@
...
+     * If the section is a soma section this function will return the soma
+     * position for all sampling positions. The soma position is assumed to
+     * be (0, 0, 0) unless the origin morphology has been transformed.
+     *
+     * @param points Normalized positions of the sample points along the
+     *        section. Values will be clampled to [0, 1] before sampling.
+     * @return The vector with all the points of the section if points is empty.
+     *         The section sampled at the given points otherwise.
+     */
+    BRAIN_API Vector4fs getSamples( const floats& points ) const;
```
eile: ```getSamples() == getSamples( Vector4fs( ))```?

brain/section.h
```diff
@@ -0,0 +1,135 @@
...
+
+    /**
+     * Return the absolute distances from the start of the section to the soma.
```
eile: distance

brain/section.h
```diff
@@ -0,0 +1,135 @@
...
+     */
+    BRAIN_API float getDistanceToSoma() const;
+
+    /**
+     * Return the absolute distances to the soma for the sample positions of
```
eile: for all sample positions?

brain/section.h
```diff
@@ -0,0 +1,135 @@
...
+     *         length is equal to the difference between the first and last
+     *         values of the list.
+     */
+    BRAIN_API floats getDistancesToSoma() const;
+
+    BRAIN_API bool hasParent() const;
+
+    BRAIN_API Section getParent() const;
+
+    BRAIN_API Sections getChildren() const;
```
eile: Immediate children, I guess?

tribal-tec:
I'm not comfortable with the impl sharing between morphology and section. Reminds too much what the SDK did wrong in my opinion. The section class is public, but is not allowed to create it standalone as it always needs to be created by morphology. Seems a step backward wrt RAII and unit tests.

Why not having the section relevant functions part of morphology?

brain/section.h
hernando: Good questions... and I don't have a good answer that's the reason for which is soma profile poly-line is returned. A user that needs an empty vector won't call this method. A user that needs the center could use 0, 0, 0 when the morphology it's known to be centered there, or compute the centroid otherwise.
I can describe what the soma ring is supposed to be.

brain/section.h
hernando: No. First it would be getSamples( floats( )). Then, getSamples( points ) returns samples at specific locations, getSamples( floats( )) would return an empty list. On the other hand getSamples() returns the sampling positions as they are.

brain/section.h
hernando: yes

brain/section.h
hernando: Yes.
How would you call the method?

hernando:
To me, the design is much more OO with the existence of a Section class. Compare the following:

    float dendriticLengh = 0;
    for( auto section : morphology.getSections({ SECTION_DENDRITE, SECTION_APICAL }))
        dendriticLength += section.getLength();
    
and

    float dendriticLengh = 0;
    for( auto id : morphology.getSectionIDs({ SECTION_DENDRITE, SECTION_APICAL }))
         dendriticLength += morphology.getSectionLength( id );

The second one looks like doing OO with C in my opinion. And since the sections keep track of the morphology safely internally, you don't need to pass the morphology reference to code operating on sections.

Your claim about being a step backward in relation to RAII and unit tests should be more concrete.
I don't understand what's the weakness that you see in this design related to RAII. Since the class doesn't have a public constructor for me there's no statement to do about RAII, it doesn't apply.
It's true that the Section class cannot be tested separately, but saying that's wrong is based on the assumption that there's a one-to-one mapping between units and classes. That's generally true, but if it had to be always the case, then why aren't they called class tests? On the other hand, it you consider that the unit is the Morphology, Section pair (which is still a small unit), then it makes sense.

Last but not least. Adding all the methods to the morphology class makes the interface look more bloated and the function names longer.

I need more solid argumentation that this is an anti-pattern or a consensus against it to revert it.

Would it make you feel more comfortable if I declare Section as a nested class of Morphology?

brain/section.h
```diff
@@ -0,0 +1,135 @@
...
+{
+class Morphology;
+}
+
+/**
+ * A class to represent a morphological section.
+ *
+ * A Section is an unbranched piece of a morphological skeleton.
+ * This class provides functions to query information about the sample points
+ * that conform the section and functions to obtain the parent and children
```
rdumusc: conform -> compose or form

brain/section.h
eile: Then I don't understand ```@return The vector with all the points of the section if points is empty.```

brain/section.h
eile: It's fine. Maybe just add doxygen to clarify.

brain/section.h
eile: Returning the center might still be valid, in particular if the caller does not have access to the original transform anymore.

brain/section.h
hernando: That's wrong. I forgot to remove that part.

brain/section.h
hernando: done

brain/section.h
hernando: I've added also documentation to getParent. I have a question there. For sections with no parent the method will throw, what should be considered a section without parent, first order sections or the soma section (I'm still not comfortable with considering the soma as a special section instead of a different concept, but many data file formats do this consideration)?

brain/section.h
hernando: That's the reason for which I don't want to make any assumption and return the points as they are. Another option, which might be more correct is to create a separate soma class and provide methods for explicitly returning the centroid, the average radius and the poly-line. If there was a separate class for this we wouldn't need to worry about defining the semantics of this operations for the soma because we can disallow creating a Section for a soma section, or at least, return empty lists to signify that the result is ill-defined.

You can still recover the translated position using the method that takes the floats vector as input, although this is not obvious at first sight.

Should the transform be stored along with the morphology? This way users are free to make their own assumptions about how the morphology is centered because all information is available.

hernando:
@tribal-tec, since you didn't answer I think I didn't convince you. Regardless of the implementation, the proposed API is easy to test in my opinion (and there are tests), and I'm trying to find a reason why the design is flawed in terms of RAII, but I don't find it. Why do you think it's better to have all the functions in the morphology class (and don't focus on the implementation for the moment)?

tribal-tec:
I haven't enough time yet to reply. I have a more philosophical problem where we define an API for something we don't really own (but use of course), w/o involving people from other teams and/or looking what was done in that regard on their end (bluepy, neurom, ...). I'm afraid it will converge to an 'SDK-light'.

The section class right now is wrapper on top of data that is owned by Morphology. That ownership is shared across any number of sections that may have been returned which can be implemented in a safe way, but still relies on the presence of  a morphology object (worst case: cf segment points this-pointer iteration because data is stored in a vector!!). If you just look on the section class, it cannot be used by itself w/o a morphology (which semantically might make sense), but then RAII is ill-formed to me as there is no resource owned or managed by that section object.

Your outlined examples wrt computing the dendrite length is outlining another thing: If a user wants to know the length of the dendrite, he/she has to compute it rather than just using a getDendriteLength() function. Of course, in the extreme that means there is >100 functions for all the different use cases, although I believe for the basic things we do with morphologies it should be reasonable small. For a high(er)-level API I would except such things rather than a convenience wrapper on top of the loader class which does not provide more real functionality. This is potentially a discussion points with other users what makes sense to them.

Last thing regarding the soma. I would not model it as section given the special handling it imposes on the section class. For neuron/cell type of morphology a soma/cell body is always present (correct me if I'm wrong), so it should be defined part of the morphology.

hernando:
Your concern about ending up with an SDK-light is reasonable, but don't be biased by assuming that everything inside the SDK is wrong, because some ideas are reusable.
Given my lack of knowledge of client code using other tools (NeuroN, BluePy), I've designed thinking of the minimal feature set needed by BRayns and RTNeuron (and not all in the case of RTNeuron), drawing ideas from BBPSDK and heavily simplifying it (no Segment, no Cross_Section, no Segments or Sections). It think the end result is neater and even more compact. It may even be more efficient because it uses fewer pointers and indirections, but I don't have hard data.

I took a quick look at NeuroM and what I've found there is even higher level (and less clear IMO, with funny things like SomaA, SomaB and SomaC as class names, but I'm also biased), I didn't find the notion of section ID for example. In BluePy they are using the BBPSDK morphology and reader and then doing some post-processing to provide bounding boxes and a transposed view on the data (the documentation is not sufficiently clear but reading the code there a Sections class that converts each section into a tuple of numpy arrays. In any case, what I see on those APIs is something that could build on top of this. Having a Section class becomes a matter of taste, I don't see the design risk.

One argument in favor of having the Section class is that it allows better information hiding. With the Section object you can forget completely about IDs, which is a detail of the data model. In the end some algorithms and data structures need to use IDs for compactness and referencing (e.g. synapses), so it's a matter of a trade-of.
But be aware that obeying users without questioning may be dangerous. According to statements I've heard many times, many end users simply want to have data file specifications and write code to read the data directly their own way, and that's wrong in so many aspects that I'm not going to extend further.

As for the comment you've made about getDendriteLength I think you've answered yourself. I don't want to take the risk of API inflation, so I haven't added anything for which I don't see an immediate use case. For example, providing the bounding volume of the morphology seems to be a very good candidate for addition.
On the other hand, I'm afraid that with NeuroM we are going to be the only consumers of this API in the end.

Regarding RAII I think you're making a logical mistake when deriving your conclusion (affirming the consequent). RAII does not imply IIRA. Of course depending on what's a resource for you, you may argue than initialization always implies resource acquisition (meaning by acquisition something more general than allocation, e.g. getting a handler to an already allocated resource). Following your line of thought you could also conclude that iterators are wrong because they don't allocate anything and are just wrappers to access the items contained by a collection.

Do we agree on having a separate class for the soma then?
If that's the case there's a small issue with Morphology::getSection(). When the requested section is the soma this method will have to throw. The only way to avoid this problem is not assigning an ID to the soma, but that's not possible with our data model.

tribal-tec:
So you'd rather take the risk of class inflation :p At this point it seems we're just discussing code anyways, so I'll play along.

For the soma, to follow your design it probably is a separate class. But it's catch-22 for the mentioned problems with getSection() with soma ID vs Section class functions that are semantically wrong/non-existent for soma sections.

hernando:
Yes, I prefer class inflation because I personally find it easier to understand many smaller pieces of information than fewer bigger ones.

In the spirit of "explicit is better than implicit" I think it's preferable to have the separate class and make getSection throw for the soma. It's still awkward, not only because of getSection not working for soma sections, but also because Morphology::getSectionIDs() is going to return an ID which can't be translated into a Section if asked for SECTION_SOMA,..
At least users wouldn't be confronted with the special semantics of the Section class for the soma and the Soma class can be very clear about the features it provides. It will require more changes in Fivox, but I think the final code is going to be clearer about how it deals with the soma.
In NeuroM I've also seen a separate class for the soma.

I'm going to make this change in a separate commit anyway, but do you think it's preferable to create a different PR for that to isolate the discussion?

hernando:
Actually, I'm thinking that having SECTION_SOMA as a SectionType is semantically wrong. The soma is a cell part, but not a section type given our definition of section. However all morphology file formats I've seen treat the soma as another section.

hernando:
I've created a separate branch for the Soma class API proposal, please comment [there](https://github.com/hernando/Brion/commit/22cbb2872f08034601d916eacd006f5282cda8cc)

rdumusc:
+1 for the separate Soma class, I agree with you about "SECTION_SOMA as a SectionType".

About the Section class debate:
- It is more OO and feels nicer to use
- It does not prevent unit testing, if tested as part of the Morphology public API
- So it should probably be a nested class, as you suggested. A brain::Morphology::Section looks more correct that a brain::Section especially given the strong interdependence (Section can't be created alone).
- The implementation sharing with detail::Morphology does feel a bit strange, but is not inherently wrong
In general the whole API really depends on the use cases, how people will want to access the data.

Last but not lease, could we *please* stop using this "detail" namespace? I thought I had managed to convince you... :-)

hernando:
The namespace detail was reintroduced due to the fact that there's an implementation class that needs to be used in two different .cpp files, not because I like it.
I can go with the nesting, and remove detail because at the moment Section becomes a member it can access Morphology::Impl without friend declarations.
That said, with nesting the most natural code distribution is to put everything in a single pair of .h and .cpp files. That makes it clear what is the unit, but makes the files long. What's your feeling about that? Separating the implementation in files is easy, but we don't have naming conventions for files implementing nested classes.

rdumusc:
OK good! (Even if the implementation must be in a separate file, I would like to name it MorphologyImpl or MorphologyData or MorphologyPrivate, over a detail namespace.)

I have the same feeling as you, the most logical way is all in one file, but that will be quite long... An option would be to name the files MorphologySection.h / MorphologySoma.cpp for example..?

hernando:
If nobody objects I'm fine with that.

... except I'm not fully sure that nesting is the best idea. I'm realizing that with nesting:
- Forward declarations of Section and Soma are not possible.
- Sections cannot be forward declared either and not making it a nested type is inconsistent.
- The Python wrapping is a bit more convoluted (unless nesting is not respected).
- The class names become longer

I will remove detail and move detail/MorphologyImpl.h to MorphologyImpl.h in any case (actually, that's what I've done in the past).

rdumusc:
Interesting, I hadn't realized that nested classes can't be forward-declared. That somehow forces us to take a clear design decision, whether Section and Soma should be usable as independent entities or are really just a part of the Morphlogy API, which won't be used separately. My gut feeling is that if they are separate entities, then they should own their own data as well (copied from Morphology).
Hard to answer without a clear understanding of the use cases.

hernando:
You have to define better "won't be used separately". I guess that independent for you means "has a public constructor",  but as of now it doesn't make sense to have one. There are no methods to modify a Section, so what would be the use for that?

In my opinion Section and Soma shouldn't be ever independent in that sense. Considering that ultimately you want Section objects to be connected in a tree, shouldn't the creation by controlled by another object (Morphology, Section or Soma depending on the case) to guarantee the invariants of the data structures? Imagine that you have a Tree structure, as desirable it may be to have a Node class, it's not sensible to let users create and connect Nodes and them to the Tree because you cannot guarantee invariants anymore (e.g. single root, no cycles, ...)
As I see it, the designed currently proposed doesn't preclude making the morphology modifiable in the future (despite the internal data structure would need lots of changes) and even in that case I doubt that having a public constructor is right... but let's not starts a BBPSDK morphology v2 discussion.

Apart from that, in order to make Section and Soma modifiable, the internal representation of the morphology would need to be transformed from the flat arrays it is into a tree structure with pointers. Since the morphology it's read-only at the moment, keeping it flat is much more efficient, and that's desirable for loading millions of them.

Even

rdumusc:
+1, I agree with what you're saying. Even if the Morphology becomes modifiable in the future, the Morphology class should maintain the full data to ensure that it can enforce the invariants (single root, no cycles, ...). Also the current design (Section keeping a pointer to the Morphology) doesn't prevent lazy loading or anything. It feels that these classes are very interdependent then. My impression is that this calls for nested classes despite the drawback of not being able to forward-declare them, but I am open for the other alternative too.
In case we decide to keep them as non-nested, I have a new suggestion: wrap Morphology, Section and Soma in a brain::neuron namespace. I am usually not a big fan of namespace inflation, but in this case it really groups things logically, makes it clear what a "Section" is, and disambiguates in prevision for future classes such as a possible glia::Morphology, etc...
With this, I am out of this discussion (Christmas holidays are coming soon), may the force be with those who finish this code review!

brain/detail/morphology.cpp
```diff
@@ -0,0 +1,249 @@
...
+{
+    const SectionRange range = getSectionRange( sectionID );
+
+    // If the section is the soma return directly the soma position.
+    if(( *types )[sectionID] == SECTION_SOMA )
+    {
+        Vector4fs result;
+        for( size_t i = 0; i != samplePoints.size(); ++i )
+            result.push_back( somaPosition );
+        return result;
```
eile: ```return Vector4fs( samplePoints.size(), somaPosition );```

brain/detail/morphology.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+    Vector4fs result;
+    result.reserve( samplePoints.size( ));
+
+    // Dealing with the degenerate case of single point sections.
+    if( range.first + 1 == range.second )
+    {
+        for( size_t i = 0; i != samplePoints.size(); ++i)
+            result.push_back(( *points )[range.first] );
+        return result;
```
eile: idem

hernando:
I started making the changes for nested classes yesterday I don't really like how it looks like, so I decided to ask for comments.

To summarize: we have reached the consensus of having separate classes for the Section and Soma concepts and disallow the creation of Section objects that refer to the Soma. Also the morphology implementation will be named Morphology::Impl (no detail namespace) and the files will be named MorphologyImpl.*.
Now the question is where all these classes should live. There are three options:

- Flat design: all classes on the brain namespace.
- Nested classes: Section and Soma nested classes of Morphology.
- Nested namespace: Morphology, Section and Soma under a nested namespace brain::neuron.

My personal preference is to go for the nested namespace. Please vote and if you have a very strong opinion against one of the options, please give arguments why.

bilgili:
I would like to understand why we need "a different" context for those parts of the morphology. And, int future how do we decide if there is needed a different context. To me the space of definitions in this case is uniform. So, I'd rather use Flat Design.

hernando:
We decide by discussing for days on Github ;)

Flat for me is also an option (as can be seen in the current PR).
The reason for nesting is that these classes are parts of a morphology than don't have (as of now) any reason to exist on their own.

Even with a mutable morphology a Section or Soma shouldn't be created separately and then added to a Morphology, but created from the Morphology or another existing Section or Soma (well, technically speaking it's possible to create the classes independently and connect them afterwards, but this freedom comes with the price of making much more complicated to guarantee the invariants and requiring the API to throw for things that in the simpler case are not possible, I haven't seen a use case that requires this so far and it goes against the design principle of preventing misuse of the API by syntax not runtime).

bilgili:
It is easy to prevent creating objects ourselves on the development level. And, letting a class to take ownership of creating them. And, I don't necessarily believe this should give them a new context. But this is an subjective opinion, rather than a fact. On development level I am not seeing, how the difference will change users look and feel to the library. My vote is still on the Flat Space.

hernando:
@tribal-tec, @eile what's your opinion?

tribal-tec:
I vote for the nested namespaces. brain::neuron::Section gives more information than brain::Section.

eile:
Ah, the boost way. I would have gone with a flat structure and split it up when/if it becomes to crowded.

Why neuron and not morphology? Shouldn't the parent class be called Neuron too then?

hernando:
I thought it was implicit that the namespace name is not decided yet. Let's decided the solution before  the details. Currently we have +2.5 for nested namespace, +1.5 for flat, (halves are my opinion). What's yours @eile?

bilgili:
1.5 - 1.5 ? ;) ( You, me and Daniel voted ? )

hernando:
And Raphael before leaving. You could also say his vote is 0.5 nested namespace 0.5 nested classes, but I think he's more in favour of the nested namespace as he proposed it.

hernando:
ping

brain/CMakeLists.txt
```diff
@@ -8,16 +8,22 @@ source_group(\\ FILES CMakeLists.txt)
 
 set(BRAIN_PUBLIC_HEADERS
   circuit.h
-  morphology.h
   types.h
+  cell/morphology.h
+  cell/section.h
+  cell/soma.h
```
eile: # All: please vote:

Option 1: As above, except s/cell/neuron/g
Option 2:
```
morphology.h
morphology/section.h
morphology/soma.h
```

## Option 1 rationale: 

All functionalities wrt morphologies contained in a single namespace. Adding glial morphology would not collide.

## Option 2 rationale: 

This option mimics nested classes cleanly using namespaces.

Symmetry with brion::Morphology and in naming: for section, it's a morphology::Section. Maybe morphology should be renamed neuron/morphology here and in brion:: (tbd when glia is introduced: aka defer Option 1).

brain/cell/soma.cpp
```diff
@@ -0,0 +1,104 @@
...
+    centroid /= float( points.size( ));
+    return centroid;
+}
+
+}
+
+Soma::Soma( Morphology::Impl* morphology )
+    : _morphology(morphology)
+{
+    _morphology->ref();
```
bilgili: is ref() thread safe ?

brain/cell/morphologyImpl.cpp
```diff
@@ -0,0 +1,252 @@
...
+#include <boost/foreach.hpp>
+
+#include <bitset>
+
+namespace brain
+{
+namespace cell
+{
+
+Morphology::Impl::Impl( const brion::Morphology& morphology )
```
bilgili: Double impl ?? Scary !

brain/cell/soma.cpp
hernando: Lunchbox says it is.

brain/cell/morphologyImpl.cpp
hernando: No, no. Morphology::Impl is the class name the other Impl is the name of the method, which is the constructor.

brain/CMakeLists.txt
jafyvilla: I like option 1 better: neuron/*.h

bilgili: I don't know much about glia to answer the question. Are morphology and gliamorphology different in terns of behaviours ? Sections etc. Or there will be a enumaration in morphology to state the type ?

hernando: As far as I know, they are different enough to make it difficult to provide base classes for both. Besides that I don't know if there's any use case that requires the base class.

rdumusc: Very hard to say.
As it is now, I have a slight preference for the option 1, as having a "neuron" namespace makes the meaning of the classes very clear: brain::neuron::Morphology,  brain::neuron::Section, brain::neuron::Soma.
If Soma is not a nested class of Morphology, because the implementation would be ugly, then I have doubts about replacing the concept with an extra namespace.
With option 2, we will still have the name collision with a future glia::Morphology, unless it is combined with option 1 later as you also suggest. But then the hierarchy becomes quite long, and I am not sure if the "morphology" namespace brings much additional information in a brain::neuron::morphology::Soma.
(The last unknown is how a future glia morphology would be implemented. I am assuming here that they will be different enough and there will not be the use for a class hierarchy between the two...)

hernando: There wouldn't be a brain::neuron::morphology::Soma in any case. It would be brain::neuron::Soma even for option 2 if it's later reworked to consider brain::glia (as Stefan says, option 2 is to defer option 1 until needed and be as consistent with brion as possible for the moment).

brain/cell/morphologyImpl.cpp
bilgili: My bad ! :)

brain/CMakeLists.txt
bilgili: I am finding it difficult to vote because of the unknowns about different cells, different morphologies. If morphology is skeleton it should not be so much different than the morphology. If they are really different they should have different folders and namespaces. If they look a lot like, there would be only type informations and then afterwards we don't even need a cell directory ( my previous flat model idea is still valid ).

Finally I am voting for "1", but it is like a coin flip where I like the HEADs a bit better and sending a "reiki" energy toward ! :)

hernando: Glial morphologies have elliptical cross sections. That alone is a big change. Then the section types are completely different also. The most correct thing would even be to create a new section type. Even more, I don't even know if a single class would be enough to represent the morphologies of all types of glial cells.

bilgili: To play the devils advocate, circles are special cases of elipses :) "a=b". If we have a section that is common between neuron and glia that should be fine.

hernando: Except that ellipses require at least 2 more floats to be fully defined (focal distance and angle), which is a waste for millions of points. But let's not discuss this yet.

brain/neuron/section.h
```diff
@@ -0,0 +1,144 @@
...
+    friend class Morphology;
+    friend class Soma;
+
+    Section( const Section& section );
+
+    ~Section();
+
+    Section& operator=( const Section& section );
+
+    bool operator==( const Section& section ) const;
```
tribal-tec: missing != operator

brain/neuron/section.h
```diff
@@ -0,0 +1,144 @@
...
+
+    SectionType getType() const;
+
+    /**
+     * Return the total length of this section in microns.
+     *
+     * If this section is a soma section the length is ill-defined and this
+     * function will return 0.
+     */
+    BRAIN_API float getLength() const;
```
tribal-tec: BRAIN_API must be added to all public functions: also for the upper operators and ctors

brain/neuron/morphology.h
```diff
@@ -91,31 +99,36 @@ class Morphology : public boost::noncopyable
...
+    BRAIN_API Sections getSections( const SectionTypes& types ) const;
+
+    /**
+     * Return the Section with the given id.
+     * @throw runtime_error if the id is out of range or the given id refers to
+     * a soma section.
+     */
+    BRAIN_API Section getSection( const uint32_t& id ) const;
+
+    BRAIN_API Soma getSoma() const;
```
tribal-tec: missing doc

brain/neuron/morphology.h
```diff
@@ -91,31 +99,36 @@ class Morphology : public boost::noncopyable
...
+
+    const Matrix4f& getTransformation() const;
```
tribal-tec: dito

brain/neuron/morphologyImpl.cpp
```diff
@@ -0,0 +1,253 @@
...
+SectionRange Morphology::Impl::getSectionRange( const uint32_t sectionID ) const
+{
+    const size_t start = ( *sections )[sectionID][0];
+    const size_t end = sectionID == sections->size() - 1 ?
+                           points->size() : ( *sections )[sectionID + 1][0];
+    return std::make_pair( start, end );
+}
+
+uint32_ts Morphology::Impl::getSectionIDs(
+    const SectionTypes& requestedTypes, bool excludeSoma ) const
```
tribal-tec: const bool

brain/neuron/morphologyImpl.h
```diff
@@ -0,0 +1,89 @@
...
+
+#include "morphology.h"
+
+#include <lunchbox/lfVector.h>
+#include <lunchbox/referenced.h>
+
+namespace brain
+{
+
+typedef std::pair< size_t, size_t > SectionRange;
```
tribal-tec: why not in neuron namespace?

brain/neuron/section.cpp
```diff
@@ -0,0 +1,132 @@
...
+}
+
+floats Section::getSampleDistancesToSoma() const
+{
+    return _morphology->getSampleDistancesToSoma( _id );
+}
+
+bool Section::hasParent() const
+{
+    int32_t parent = ( *_morphology->sections )[_id][1];
```
tribal-tec: const int32_t

brain/neuron/section.cpp
```diff
@@ -0,0 +1,132 @@
...
+    return parent != -1 && uint32_t(parent) != _morphology->somaSection;
+}
+Section Section::getParent() const
+{
+    int32_t parent = ( *_morphology->sections )[_id][1];
```
tribal-tec: dito

brain/neuron/section.cpp
```diff
@@ -0,0 +1,132 @@
...
+    if( parent == -1 || uint32_t(parent) == _morphology->somaSection )
+        LBTHROW( std::runtime_error( "Cannot access parent section" ));
+    return Section( parent, _morphology );
+}
+
+Sections Section::getChildren() const
+{
+    const uint32_ts& children = _morphology->getChildren( _id );
+    Sections result;
```
tribal-tec: reserve?

brain/neuron/section.h
```diff
@@ -0,0 +1,144 @@
...
+    ~Section();
+
+    Section& operator=( const Section& section );
+
+    bool operator==( const Section& section ) const;
+
+    /** Return the ID of this section. */
+    uint32_t getID() const;
+
+    SectionType getType() const;
```
tribal-tec: missing doc

brain/neuron/soma.h
```diff
@@ -0,0 +1,87 @@
...
+ * coordinate is not 0 for all the points).
+ * This class can also be used for both descriptions as well as somas simply
+ * approximated as spheres.
+ *
+ * The coordinates system used by a soma will be in the same as the
+ * brain::Morphology from where it comes.
+ *
+ * @version unstable
+ */
+class Soma
```
tribal-tec: missing doc for some functions, operator !=, BRAIN_API

hernando: There's no operator== either. Should I add both?

hernando: The only functions with missing documentation are the copy constructor and the assigment operator. Do you really want me to add it?

tribal-tec: Oh, now it looks different... It's fine then.

hernando:
All comments have been addressed, I don't know why Github is not obsoleting them.

PR #36, opened 2015-12-16
Fix doxygen namespace generation for brain

eile:


hernando:
Moving the file lists form files.cmake to CMakeLists.txt is something that Daniel suggested before, but I didn't change it because Brion has a files.cmake. For consistency it should be the same in both libraries.

eile:
(sigh) incremental vs. global changes again. I'll do it for Brain to, although imo this is not a reason to block the PR.

eile:
Just saw your PR.

eile:
+1 then?

hernando:
Too quick complaining ;)

eile:
Asking! ;)

hernando:
Sometimes I think that reviewers comments could be addressed with PRs on the original PR like in this case. This will save time because explaining the requested changes sometimes is even longer than directly doing them (e.g. typos). We should discuss this approach.

PR #37, opened 2015-12-17
Added support for synapse nrn_extra.h5 files.

hernando:


tribal-tec:
Is there anything we can/should document what this file is? From the changelog, it's not clear.

hernando:
I don't know how stable it is. In principle this should be documented sooner or later in the repositories Jean-Denis leads the task of creating.

The other files are not particularly documented either. What I can do is write a high level description of all the synapse files in the class documentation, without describing the precise layout.

tribal-tec:
At least the one attribute (one file for one attribute???) should be mentioned somewhere. The 'real' synapse file has at least all the enums where you get an idea what's inside. For extra, I have no idea :)

hernando:
This new file reflects how badly are data being managed in some parts of the toolchain. I don't like it at all, but I guess it's take it or leave it. It's very frustrating to be the last ones on the consumer chain, when end up eating all the poison, all this things would look quite differently if we were the producers.

It would have been much more sensible to add the attribute to the nrn.h5 file, but guess what?, that was breaking some of the tools because they access the .h5 files directly of course.

eile:
@hernando ack, although I'm not sure how "different" would look if we were guilty of it.

@delalond: Why was this on attribute written in a new file? Surely that can't be a long term solution?

hernando:
What are we going to do with this review? The related review on BBPSDK has been merged.

eile:
+1

PR #40, opened 2016-01-28
Spec for MPI-enabled compartment report write

eile:
Pretty version at https://github.com/eile/Brion/blob/master/doc/feature/compartmentReportMPI.md


eile:
@tribal-tec FYI

PR #44, opened 2016-02-29
Specification document for a new SpikeReport API.

hernando:


doc/feature/spikeReport.md
```diff
@@ -0,0 +1,417 @@
...
+* When data for a time interval is requested, the time intervals are always
+  closed on the left and open on the right.
+
+## API
+
+The classes presented below are not thread-safe except for const access.
+
+### Spikes container
+
+\code{.cpp}
```
eile: doesn't render correctly. Just outdent all by four spaces, and remove \code

hernando: I made it to be rendered correctly by Doxygen. How are you looking at it?

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,453 @@
...
+    SpikeReport destination( "url2", brion::MODE_WRITE )
+
+    float current = source.getStartTime();
+    float width = 10;
+    // If we have an empty file report, current == UNDEFINED_TIMESTAMP, and then
+    // the loop is not entered.
+    for( ; current + width < source.getEndTime( ); current += width )
+    {
+        const Spikes spikes = source.getSpikes( current, current + width );
+        last = spikes.getEndTime();
```
bilgili: last is used somewhere ?

hernando: No, it's a copy-paste from the other example.

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,453 @@
...
+
+
+### 1: Why do we need a Spikes container
+
+_Resolved: No_
+
+There are arguments against and in favour of it.
+
+Against:
+* It adds unnecessary complexity to the implementation.
```
bilgili: Dont forget Mike : @mgeplf :)

mgeplf: Thanks!

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,450 @@
...
+    public:
+        SpikeReport( const URI& uri, int mode );
+
+        /*
+         * Open a report in read mode with a subset selections.
+         *
+         * @param uri
+         * @param subset The set of gids to be reported.
+         */
+        SpikeReport( const URI& uri, uint32_ts subset );
```
mgeplf: const uint32_ts& ?

is there a way to query which gids are contained in the report? (I have probably missed it)

hernando: yes, const uint32_ts%

So far we make no assumptions about the target contained in the report. Having a function to get that information would require changing the file formats because currently you can't know which cells are there unless you load the whole data. And even in that case, you will just be guessing which is the reported target because silent cells won't be there.

eile: @mgeplf :  'type + s' is our convention for 'std::vector< type >'. works better with class types then builtins

eile: @hernando : So far we use GIDSet not uint32_ts/GIDVector throughout brion - I think we should keep this consistent even if it's less optimal.

eile: Document contract if gids requested are not in the report. Simply never reported? Throws?

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,450 @@
...
+
+brion::SpikeFile?
+brion::SpikeStream?
+
+## Issues
+
+
+### 1: Why do we need a Spikes container
+
+_Resolved: No_
```
eile: :)

I guess the question should be 'Do we need a Spikes container?'

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,450 @@
...
+         * @see SpikesReport::getSpikes.
+         */
+        float getStartTime() const;
+
+        /*
+         * Get the end of the time window in milliseconds.
+         *
+         * UNDEFINED_TIMESTAMP if the container cannot be referred to any time
+         * window and it is empty. Otherwise, a time greater than the highest
+         * spike time in the container.
```
eile: strictly greater? Why?

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,450 @@
...
+         * @throws std::runtime_eror if the report was opened in read mode.
+         * @throws std::runtime_eror when writing to a stream if the spikes are
+         *         not sorted or any timestamp is <= to the highest timestamp
+         *         written by previous calls. This is required to enforce the
+         *         post-conditions of getSpikes in the receiving end.
+         */
+        void writeSpikes( const const std::vector< float, uint32_t >& spikes );
+
+        /* Overload of the method above provided for convenience */
+        void writeSpikes( const Spikes& spikes );
```
eile: read() and 2x write() for method names

eile: or get() / set()

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,450 @@
...
+         * last known timestamp.
+         *
+         * This function is thread-safe with getSpikes as it ensures that any
+         * thread waiting there with an infinite timeout will eventually return.
+         *
+         * Implicitly called by the destructor, (but calling the destructor and
+         * getSpikes at the same time is not safe!). Calling any other function
+         * after the report has been closed has undefined behavior.
+         */
+        void close();
```
eile: Use case? flush() sufficient? use dtor instead?

eile: close() actually violates RAII, since it leaves you with an invalid object.

doc/feature/spikeReport.md
mgeplf: @eile Yeah, I gathered that from the typedef, I was just pointing out it may as well be be const-ref.

hernando: I forgot about GIDSet. Yes, it should be that type.

For missing GIDs I think the right solution is to ignore them without any notice. Otherwise we need to know which are the valid GIDs at source, which is not unfeasible, but requires either loading all the data or changing the format for files and becomes quite complicated for streams.

doc/feature/spikeReport.md
hernando: Consistency with the specification of the interval passed to getSpikes as open on the right, I think it's mentioned in the issues (we may add it otherwise). If you call getSpikes using the values returned by this container you will get back the same content again. Otherwise it would be shrinking. I think it's important to always use the same time of intervals ([start, end)), because it makes reasoning about interval algebra easier. This specification also prevents having things like [t, t) which is wrong, and [t, t], which is not only inconsistent with getSpikes() but may also be problematic because it has 0 duration but still may contain events. In the specification of getSpikes it says (maybe not so clearly) that the end time must be either undefined or greater than the start time.

doc/feature/spikeReport.md
hernando: I prefer read/write. setSpikes is not very clear about IO being involved.

doc/feature/spikeReport.md
hernando: close() is intented to unlock any thread that may be potentially blocked with an infinite timeout in getSpikes(). Without this function the only alternative available for multithreaded programs that want to exit orderly is to always use a timeout in getSpikes, which is not very nice.

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,450 @@
...
+As a side-effect this requires Spikes to be copyable and the copies need to be
+efficient, but thanks to issues #1 and #2 the implementation can provide
+cheap copies by sharing.
+
+### 4: Do you need a method to erase portions of the report?
+
+_Resolved: Partially_
+
+We don't have a use case for this yet, but it looks necessary for out-of-core
+processing of very large reports. The mail goal for this method in the original
```
eile: typo mail

doc/feature/spikeReport.md
eile: ack for const ref. missed that.

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,450 @@
...
+class was to clear the internal caching. Clearing the report can be very
+complex without requiring duplication of data. The compromise solution is to
+provide a method to clear the cache enterily. The implementation can easily
```
eile: type enterily

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,450 @@
...
+keep Spike containers returned by getSpikes valid with smart pointers.
```
eile: API should be consistent with CompartmentReport. This one has setBufferSize() - so either should be SpikeReport::setBufferSize( const float time ) or be changed/removed in CompRep.

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,450 @@
...
+
+### 6: For stream based reports should we block on getSpikes() or in the iterator ++ operator.
```
eile: We're missing an issue 5.

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,450 @@
...
+
+It seems easier for the implementation to block on getSpikes than on
+the operator increment. Blocking in the container requires synchronization
+primitives to ensure const-correctness that wouldn't be otherwise needed.
+Locking at the iterator increment requires having a separate function to
+indicate the timeout, which is awkward, or not having a timeout at all for
+an equivalent to waitUntil and also add the current functions to know until
+which time is it possible to advance without blocking.
+
+### 7: Why getStartTime() retuns 0 for stream based reports, but it can be undefined for empty file reports?
```
eile: type retuns

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,450 @@
...
+
+_Resolved: No_
+
+Seems reasonable, but it's open to discussion.
```
eile: Why can't stream reports have the same behaviour as file, i.e. return UNDEFINED when nothing was received and the time of the first spike received (and still buffered)?

doc/feature/spikeReport.md
hernando: you made the same typo :D

doc/feature/spikeReport.md
hernando: Difficult to say. In compartment reports the size of the data in a time window is known. With the spike report you don't know. A window of [0, 10s] could be to big or too small it terms of total data. At the same time, in the CompartmentReport you read frame by frame, and your buffer is clearly a set of frames. What does the buffering mean in a SpikeReport if I set the buffer size to 42 and then I do getSpikes(10, 100)?

I don't consider the consistency between the two report APIs so critical because they are fundamentally different. A compartment report is a function f: R, N -> R (time x offset -> value), a spike report is f: R, R -> set((R, N)).

doc/feature/spikeReport.md
hernando: In file based reports an empty report has UNDEFINED start and end times and that can't change over the course of time, so the client knows that there isn't anything to do with that. In stream reports we need to differentiate very clearly a finished stream from an ongoing one. If no spike data has been received and we make start time to return UNDEFINED that can be confused with the empty file report (unless we add isStream(), but you don't like that, do you?)

doc/feature/spikeReport.md
hernando: outdated commit?

eile:
Discussion notes:
* readSpikes/writeSpikes -> read/write.
* No custom container. Use ```vector< time, gid >``` or ```map< time, GIDs >```. Not sure which is preferable. Second option could have a ```endTime, empty``` entry to indicate up to which time spikes have been read.
* read( *, UNDEFINED ) should always block until all data is read, for both file and stream. Use timeout or concrete endtime if other behaviour is needed
* iterator interface could be provided, always non-blocking. would need ```advance( time )``` like method (aka read w/out return copy value)
* clear() use case?

hernando:
I don't agree with the third one.
There should be a way to wait for new data that is guaranteed to return in a finite but undefined amount of time for reports that are making progress and are infinite without having to specify a timeout. The example is very clear, copying a stream to somewhere else, which is exactly the first example provided.
First I don't want to specify a timeout for such a simple thing because choosing the right timeout is very use case sensitive and depending on the case the client code can become a soft version of an active wait loop.
Second, I can't use getEndTime to know until where I can read unless it's made non const (so it's allowed to return different values each time is called without calling any other non-const function in between), which may be problematic (I remember we have already have issues with a similar thing). And even is I could use getEndTime I'm not sure that the loop wouldn't end up being a active wait.

An alternative to the proposed spec is adding a wait method that waits for any event to occur. After waiting you can use a const getEndTime to read until the end of the available data, but that also means that we need to add a new function to detect the end of the stream which currently is done with getEndTime. And that's two new stream specific functions in the API.

hernando:
Any change proposed should consider getEndTime, getStartTime, and read altogether (and even the container return type) because I don't think it's possible to change the current specification of any of them separately without breaking the whole design.

hernando:
I'd added an example use case for clear(). You may argue that this is not needed if read returns a std container by value and the data is always read from the source. However in stream reports that won't work because once you throw away the data there's no way to get it back. So you need to store it internally until the user wants to get rid of it (which could be by creating a new SpikeReport, but that implies that some spikes could be missed).

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,476 @@
...
+## Examples
+
+### Copying a spike report from one URL to another
+
+    SpikeReport source( "url1", brion::MODE_READ )
+    SpikeReport destination( "url2", brion::MODE_WRITE )
+
+    float last = 0;
+    // In file based reports this will iterates just once.
+    while( source.getEndTime() == brion::UNDEFINED_TIMESTAMP )
```
bilgili: how about having an END_OF_REPORT instead of UNDEFINED. That is not informative at all. It can underneath be undefined but for a user using the API, it is confusing.

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,476 @@
...
+    {
+        const Spikes spikes = source.read( last, brion::UNDEFINED_TIMESTAMP );
```
bilgili: How about ? It is more similar to std::stream etc. I dont want to reshuffle everything but, as a user it feels much more convenient.

const float time_t = source.getEndTime();
while( !source.endOfReport())
{
Spikes spikes;
const time_t timeDiff  = source.read( spikes, begin, brion::END_OF_REPORT );
begin += time ;
}

eile:
Keyword: Discussion notes. ;)

Point 3: By the same logic, if read( *, INFINITE) is called on a file report, it should return each spike (or atomically readable time point) individually.

eile:
Re clear: I wanted to postpone this feature to when we have a use case. Eg: Why clear() and not setBufferSize() which is applied on read()?

hernando:
I was starting the discussion by giving my opinion on 3.

That's not the same logic (indeed I don't know to which logic to you refer). I said: "There should be a way to wait for new data that is guaranteed to return in a finite but undefined amount of time for reports that are making progress and are infinite without having to specify a timeout.", and I decided to use read(start, UNDEFINED) for that purpose. When I say new data that could be one spike, or all, I haven't stated anything about how much.
Then, the spec says: "... this function will return all the spikes available with timestamp >= startTime", the keyword is _available_. With this wording I don't see the inconsistency between file reports and stream reports. In file based reports all the data is always available, so the function returns all until the last spike. In stream reports your endTime is already undefined (unless the report has been closed), so if I decide to not make the function wait until the report is closed but allow it to return earlier, I'm not making it inconsistent with the statement about "all the spikes available" or how the file based reports work.

If you read carefully, the debatable part is about what read(UNDEFINED, UNDEFINED) means. I assumed that being able to get all the available data without blocking at all may be necessary in some cases (e.g. know the time window of the data received so far without blocking, note again that getEndTime cannot be used for this in stream reports not closed yet, and also note that returning a std container will force us to add new functions to avoid copying data when we want to know this information) and I decided to use this special combination for that purpose. Doing r.read(r.getStartTime(), r.getEndTime()) works as a blocking call for stream reports in the general case because getStartTime() returns 0 when no data has been received so far (also debatable), but now I've realized that getting all the data without blocking can be explicitly requested with a 0 timeout.
Then if no data has been received, getStartTime() should return UNDEFINED also in stream based reports, and once the report got data, return the time of the first spike. That way the specification is consistent for both file and stream based reports. However, the code of the second example would be wrong because it does {{ float current = source.getStartTime(); }} and that doesn't work for stream based reports. 

I think I'll remove the particularity about read(UNDEFINED, UNDEFINED), but I'm not so sure changing the spec for getStartTime() is a good idea.

hernando:
I'm reading the specification and and I'm realizing why I made read(UNDEFINED, UNDEFINED) a special case in stream based reports. The question is the following, if you have a stream from which you have received a few spikes and you call read(UNDEFINED, UNDEFINED), it's clear that the specification for read(t, UNDEFINED) is not accurate, because "_the call will block until an event with t > startTime arrives_" is meaningless if t == UNDEFINED.

hernando:
Sorry, I wanted to close the comment editor for not submitting a comment and pressed the wrong button.

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,289 @@
+\Redesign of the SpikeReport and Spikes container
```
eile: s/\//

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,289 @@
...
+         *        actually part of the report will be silently ignored.
+         */
+        SpikeReport( const URI& uri, const GIDSet& subset );
+
+        /**
+         * Release all resources.
+         * Any threads waiting on futures will return immediately and calling
+         * get will throw a runtime_exception.
+         */
+        ~SPikeReport();
```
eile: ~SpikeReport

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,289 @@
...
+         * @param subset The set of gids to be reported. This set should be
+         *        understood as a filter: any included GID which is not
+         *        actually part of the report will be silently ignored.
+         */
+        SpikeReport( const URI& uri, const GIDSet& subset );
+
+        /**
+         * Release all resources.
+         * Any threads waiting on futures will return immediately and calling
+         * get will throw a runtime_exception.
```
eile: std::runtime_exception

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,289 @@
...
+        /**
+         * Read all available spikes.
+         *
+         * This is a convenience function that behaves like:
+         * read( -inf ) is getCurrentTime() is undefined and
+         * read( getCurrentTime( )) otherwise.
+         */
+        boost::future< Spikes > read();
+
+        enum StateFlags { ENDED = 1, FAILED = 2 };
```
eile: Add OK = 0?

hernando: Given that StateFlags is a bit set, I'm not sure if that's a good idea. With the other bits an operation like if( state & BIT ) is enough to check if the bit is on, but with OK that code would be wrong.

bilgili: why bits ? Can there be two states ?

hernando: It would be three states then. Whatever you prefer, I don't have a strong opinion about this.

bilgili: I mean two states at the same time.

eile:
How does one cancel a future to perform a different operation?

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,338 @@
...
+         * right. That means that upon completion of a read or write operation
+         * no spike read or written may have a timestamp >= getCurrentTime().
+         */
+        float getCurrentTime() const;
+
+        /**
+         * Read spikes until getCurrentTime becomes > min, the end of the source
+         * is reached or the report closed.
+         *
+         * This function fetches as much data as possible from the source.
```
eile: s/possible/available/?

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,338 @@
...
+         *
+         * The time inverval to which getCurrentTime refers is open on the
+         * right. That means that upon completion of a read or write operation
+         * no spike read or written may have a timestamp >= getCurrentTime().
+         */
+        float getCurrentTime() const;
+
+        /**
+         * Read spikes until getCurrentTime becomes > min, the end of the source
+         * is reached or the report closed.
```
eile: or the end of the report is reached?

(closed == end of report)

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,338 @@
...
+        /**
+         * @return the end time of the latest complete read/write operation or
+         *         UNDEFINED_TIMESTAMP if no operation has been issued.
+         *
+         * Read operations are deemed as complete when the returned future is
+         * ready.
+         *
+         * The time inverval to which getCurrentTime refers is open on the
+         * right. That means that upon completion of a read or write operation
+         * no spike read or written may have a timestamp >= getCurrentTime().
```
eile: what is returned on 'Report( "stream" ).getCurrentTime()'?

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,338 @@
...
+         * @param min The minimum end time until which spikes will be read. If
+         *        UNDEFINED_TIMESTAMP, it will be considered as infinite.
+         */
+        boost::future< Spikes > read( float min );
+
+        /**
+         * Read all available spikes.
+         *
+         * This is a convenience function that behaves like read( -inf ) if
+         * getCurrentTime() is undefined and read( getCurrentTime( )) otherwise.
```
eile: Should getCurrentTime() return -inf instead of undefined? Would simplify doc and semantics.

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,338 @@
...
+         *
+         * Precondition: max > getCurrentTime()
+         * Postcondition: If r.getState() == 0, then r.getCurrentTime() == max
+         *
+         * WIP
+         * - Why no min parameter, because it doesn´t make sense
+         * - hmmmm.. if read( 1000 ) times out, and then readUntil( 10 ) is
+         * called and succeeds, we have to move data from the container that
+         * was supposed to be returned in the first operation (and was kept
+         * for the next attempt) into the container for the second.x
```
eile: @sa forward() ?

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,338 @@
...
+         */
+        boost::future< Spikes > readUntil( float max );
+
+        /**
+         * Forward to a given absolute timestamp.
+         *
+         * In read mode this means skipping data forward until the timestamp
+         * is made current. In write mode for streams, subscribers are notified
+         * about the progress.
```
eile: In write mode consumers are notified about the progress.

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,338 @@
...
+         *  - f be the returned future by r.forward(toTimestamp)
+         * Then:
+         *  - After f.wait() returns r.getCurrentTime() == toTimestamp.
+         *  - The postconditions of read operations imply that this function
+         *    throws away the data previous to toTimestamp (or avoids reading
+         *    it at all if possible).
+         */
+        boost::future< void > forward( float toTimestamp );
+
+        enum StateFlags { ENDED = 1, FAILED = 2 };
```
eile: Still no OK=0?

hernando: As I said, if this is a bit mask OK=0 doesn't make sense to me. If this is a State enum then I should be added. Probably it's better, the question is whether ENDED|FAILED is a state that makes sense.

doc/feature/spikeReport.md
hernando: That's certainly a possibility that I considered. It's convenient, but on the other hand I'm not sure if it's logically correct.
What you are suggesting is to do that change and remove this function so r.read(r.getCurrentTime()) always works?

doc/feature/spikeReport.md
hernando: better

hernando:
Regarding the cancellation of operations, I never considered that should be possible. I think it can be really tricky both to define what's the object state afterwards and implement it properly. Besides that, I don't see a use case for that complication except when the report is destroyed, which we already consider. In that case boost::future can be used to notify that the operation was canceled.

doc/feature/spikeReport.md
eile: Right now 0 is implicitly OK. This seems wrong as well. Naming is wrong in any case, since a state is not a bit mask to me.

doc/feature/spikeReport.md
eile: Yes.

UNDEFINED can be set to -inf, and read( UNDEFINED ) reads all available spikes (since you did not define a min).

doc/feature/spikeReport.md
hernando: Well, the name was not State, but StateFlags. In any case I've already changed it to not be a bitmask.

eile:
Re cancel: Not sure we need to consider now, but the user might cancel it and pick a 'closer' time.

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,333 @@
...
+         * Closes the report.
+         *
+         * In read mode this call forces the state to be changed to ENDED.
+         * Any call waiting in any future will be unblocked as if there was
+         * no more data to be read.
+         *
+         * In write mode with stream reports this signals the end of the
+         * report to the consumers.
+         */
+        void close();
```
tribal-tec: Explicit close is needed or dtor would be enough? Can a closed report be used again for even a different operation?

hernando: Before adding the futures the explicit close was needed. Now it's possible that we don't need it at all. However that means that the thread that will interrupt the blocked threads needs to be able to destroy the report and I'm not sure if that's a very strong requirement. I need to check existing code for a better answer.

doc/feature/spikeReport.md
hernando: Since the description already says that this function "fetches all the available data from the source" I think it's enough to modify "If UNDEFINED_TIMESTAMP, it will be considered as infinite." to say "it will be considered as - infinite" and leave getCurrentTime() as is.

doc/feature/spikeReport.md
hernando: A closed report becomes unusable.

hernando:
I prefer postponing the discussion of cancel until we find the need for it. It should be easy to add at the API level (since the operations cannot overlap I simple cancel() method could do the job).

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,327 @@
...
+## API
+
+Thread-safey is only guaranteed for const access.
+
+### Helper types
+
+    struct Spike
+    {
+        Spike( float timestamp_, uint32_t gid_ );
+        float timestamp;
```
eile: why timestamp over time?

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,327 @@
...
+    public:
+        SpikeReport( const URI& uri, int mode );
+
+        /*
+         * Open a report in read mode with a subset selection.
+         *
+         * @param uri
+         * @param subset The set of gids to be reported. This set should be
+         *        understood as a filter: any included GID which is not
+         *        actually part of the report will be silently ignored.
```
eile: empty == all? If yes, document. I prefer these details in the doxygen body rather then the param description.

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,327 @@
...
+         * If no error occurs and the end is not reached, the min value passed
+         * is also guaranteed to be inside the time window of the data returned.
+         *
+         * Preconditions:
+         * - r.getState() is OK or ENDED.
+         * - The report was open in read mode.
+         * - There is no previous read or forward operation with a pending
+         *   future.
+         * - min >= getCurrentTime() or UNDEFINED_TIMESTAMP.
+         * Throw std::logic_error if one of the preconditions is not fulfilled.
```
eile: iirc @throw exists in doxygen

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,327 @@
...
+         *
+         * Postconditions:
+         * Let:
+         *  - r be the SpikeReport
+         *  - f be the returned future by r.read(min)
+         *  - and s = r.getCurrentTime() before read is called or -inf if
+         *    undefined:
```
eile: - and s = r.getCurrentTime() before read is called (s = -inf for undefined)?

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,327 @@
...
+         * After f.wait() returns the following are all true:
+         *  - r.getCurrentTime() >= s
+         *  - If r.getState() == OK, then r.getCurrentTime() > min
+         *  - For each spike timestamp t_s in f.get(),
+         *    s <= t_s < r.getCurrentTime() (Note this could collapse to an
+         *    empty interval is r.getState() != OK)
```
eile: r/is/if/?

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,327 @@
...
+         * Read spikes until getCurrentTime() == max, the end of the report is
+         * reached or the report closed.
+         *
+         * This function is very similar to the normal read, but instead of
+         * specifying a lower bound of getCurrentTime at return, it specifies
+         * a requested strict value. The preconditions and postconditions
+         * are the same except for those involving the parameter min which
+         * become:
+         *
+         * Precondition: max > getCurrentTime()
```
eile: @throw?

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,327 @@
...
+
+        /**
+         * Forward to a given absolute timestamp.
+         *
+         * In read mode this means skipping data forward until the timestamp
+         * is made current. In write mode for streams, consumers are notified
+         * about the progress.
+         *
+         * Preconditions:
+         * - toTimestamp >= getCurrentTime()
```
eile: @throw?

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,327 @@
...
+
+        enum State { OK = 0, ENDED = 1, FAILED = 2 };
+        /** @return The state after the last completed operation. */
+        State getState() const;
+
+       /**
+         * Write the given spikes to the output.
+         *
+         * Upon return getCurrenTime() is the greatest of all the spike times
+         * plus an epsilon..
```
eile: ..

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,327 @@
...
+         *
+         * @param spikes A sorted collection of spikes. For every spike, its
+         *        timestamp t must be >= getCurrentTime().
```
eile: @throw?

doc/feature/spikeReport.md
eile: I vote for implementing iff we need it.

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,327 @@
...
+
+### Copying a spike report from one URL to another without any timeout.
+
+    SpikeReport input( "url1", brion::MODE_READ )
+    SpikeReport output( "url2", brion::MODE_WRITE )
+
+    while( input.getState() == OK )
+        output.write( input.read().get() ))
+
+    if (input .getState() & SpikeReport::FAILED)
```
eile: s/&/==/

eile: same for output.

doc/feature/spikeReport.md
```diff
@@ -0,0 +1,327 @@
...
+    // advance until the current timestamp.
+
+### Copying a spike report from one URL to another with a read timeout.
+
+    SpikeReport intput( "url1", brion::MODE_READ )
+    SpikeReport output( "url2", brion::MODE_WRITE )
+
+    while( input.getState() == OK )
+    {
+        auto future = intput.read();
```
eile: intput

doc/feature/spikeReport.md
mgeplf: What is the expected range of timestamp (0 - 10, 0 - 100, 0 -1000)?  Depending on how large they become, the precision may not be precise enough.

eile: From what I can see in out.dat, floats are saved.

hernando: No particular reason for using timestamp instead of time.

I don't think precision is going to be a problem in the general case, but I think it's a problem for reproducibility if decimal representations are used somewhere in the pipeline (e.g. 0.1 does not have an exact floating point equivalent, no matter how much precision is used). However, the problem is not solved by using double instead of float.

hernando: With float precision and assuming  that 0.1 ms is a reasonable resolution we can go up to 100 s of biological time. That's definitely short for plasticity simulations. Using an unsigned integer and microseconds gives us enough for 1.2 h of biological time. Using a double is better but double the storage space.

doc/feature/spikeReport.md
hernando: Empty is empty, I think that's how it should be understood as long as that particular case is not mentioned.

doc/feature/spikeReport.md
hernando: Yes, but it breaks the flow of text. I don't think the rendering is going to look correct if I add it.

doc/feature/spikeReport.md
hernando: May not be needed at all. I'll remove it.

doc/feature/spikeReport.md
eile: So if I want all, I need to use the other ctor?

doc/feature/spikeReport.md
mgeplf: It's something to consider, imo.  With 100s, I get a precision of 0.0078125ms.  I'm probably doing my calculations wrong.

After 1hr, the minimum precision w/ a float (where the base unit is ms), the resolution falls to 0.25ms.  Not sure what people expect here.

doc/feature/spikeReport.md
hernando: done

doc/feature/spikeReport.md
hernando: In Neuron, the simulation resolution for spike time reporting is 0.0125, but interestingly, it drifts away as it advances because the simulation time is increased using an accumulator instead of doing # iteration times dt.
Jim and I have created a Confluence page to discuss about it: https://bbpteam.epfl.ch/project/spaces/display/MEETBUILDSIM/Spike+time+precision. Feel free to edit or comment.

doc/feature/spikeReport.md
hernando: Yes

doc/feature/spikeReport.md
hernando: done

doc/feature/spikeReport.md
hernando: done

PR #45, opened 2016-03-02
Made targets optional

favreau:


brain/circuit.cpp
```diff
@@ -215,6 +230,12 @@ class MVD3 : public Circuit::Impl
         , _circuit( config.getCircuitSource().getPath( ))
     {}
 
+    size_t getNumNeurons() const final
+    {
+        // Dirty cast until ::MVD3::MVD3File::getNbNeuron becomes const
```
tribal-tec: you opened a PR for that in MVDTool?

brain/circuit.h
```diff
@@ -84,6 +84,8 @@ class Circuit : public boost::noncopyable
     BRAIN_API Vector3fs getPositions( const GIDSet& gids ) const;
     /** @return The local to world transformations of the given cells. */
     BRAIN_API Matrix4fs getTransforms( const GIDSet& gids ) const;
+    /** @return The number of neurons in the circuit. */
+    size_t getNumNeurons() const;
```
tribal-tec: missing BRAIN_API

brain/circuit.cpp
```diff
@@ -103,10 +103,20 @@ class Circuit::Impl
...
-        return brion::Target::parse(
-            _targetParsers, target.empty() ? _circuitTarget : target );
+        if( target.empty( ))
+        {
+            GIDSet gids;
+            for( size_t i = 0; i < getNumNeurons(); ++i )
+                gids.insert( i + 1 );
+            return gids;
+        }
+        else
```
tribal-tec: else not needed

tribal-tec:
Looks good! Could you also add a unit test for that?

brion/blueConfig.cpp
```diff
@@ -239,11 +239,17 @@ brion::Targets BlueConfig::getTargets() const
...
-    targets.push_back( brion::Target(
-        get( brion::CONFIGSECTION_RUN, run, BLUECONFIG_NRN_PATH_KEY  ) +
-        CIRCUIT_TARGET_FILE  ));
-    targets.push_back( brion::Target(
-        get( brion::CONFIGSECTION_RUN, run, BLUECONFIG_TARGET_FILE_KEY )));
+
+    const std::string& nrnPath =
+        get( brion::CONFIGSECTION_RUN, run, BLUECONFIG_NRN_PATH_KEY );
+    if( !nrnPath.empty( ))
+        targets.push_back( brion::Target( nrnPath + CIRCUIT_TARGET_FILE ));
```
hernando: What is nrnPath doesn't have a / at the end?

brain/circuit.cpp
hernando: You have radically changed the semantics of what an empty target is, just look at the code and you'll realize it.
What's your intention here?

hernando:
-1

brain/circuit.cpp
favreau: I spoke to Adrien and he said this was already working on it. Still, I guess I can create an issue to make sure this does not get forgotten.

brain/circuit.cpp
favreau: We had a discussion this morning and agreed with that solution, but I guess you should have been in the loop. It seemed ok to assume that specifying no target would mean getting all the GIDs. What you are proposing as an alternative was also mentioned during the discussion and can clearly be done without modifying Brion. To give you a bit of background, the idea is that we must be able to open and visualize a circuit even if it does not contain any target yet. From what you are saying, modifying Brion is the wrong idea?

brain/circuit.cpp
eile: Do it for this one function and submit a PR.

brain/circuit.cpp
favreau: OK, following our discussion, there are actually 3 cases: A target is specified and we use it, a target is not specified and we may either return all the GIDs or use the target that is specified in the config file. In the end, we decided to leave this getGIDs as it was, and let the calling application deal with what to do whenever no GIDs are returned. Still, the getNumNeurons method is needed for dealing with such cases.

brain/circuit.cpp
favreau: Done

brain/circuit.h
favreau: Fixed

brion/blueConfig.cpp
favreau: I added a '/' in any case. Having 2 won't hurt anyway.

brain/circuit.cpp
```diff
@@ -135,6 +137,11 @@ class MVD2 : public Circuit::Impl
         , _circuit( config.getCircuitSource().getPath( ))
     {}
 
+    size_t getNumNeurons() const
```
tribal-tec: final

tribal-tec:
Changelog update?

tribal-tec:
And add a test for the new function

favreau:
Do not review yet

favreau:
Ready for review

favreau:
ping :-)

brain/circuit.cpp
tribal-tec: Which is the PR?

doc/Changelog.md
```diff
@@ -3,6 +3,8 @@ Changelog {#Changelog}
 
 # git master
 
+* [45](https://github.com/BlueBrain/Brion/pull/45):
+  Made targets optional
```
tribal-tec: A bit more elaborate because it's for the user: brain::Circuit::getIDs() does not require targets anymore or something similar

brain/circuit.cpp
favreau: https://github.com/BlueBrain/MVDTool/issues/10

doc/Changelog.md
hernando: Don't forget that you've also added the getNumNeurons() function in this commit.

brain/circuit.cpp
tribal-tec: That's an issue, not a pull request :) I thought the idea was you fix this one line in MVDTool to avoid the cast in Brion.

favreau: Indeed :-) I understood Adrien was on it last week so I let him do. I'll check with him what the status currently is.

doc/Changelog.md
favreau: Thanks for the reminder.

favreau:
Do not review yet please

favreau:
Ok for review

brion/target.cpp
```diff
@@ -147,11 +147,12 @@ const Strings& Target::get( const std::string& name ) const
 
 namespace
 {
-void _parse( const Targets& targets, const std::string& name, GIDSet& gids )
+void _parse( const Targets& targets, const std::string& name, GIDSet& gids, bool& found )
```
tribal-tec: 21st century dictates return value, not out parameter :p

brion/target.cpp
```diff
@@ -147,11 +147,12 @@ const Strings& Target::get( const std::string& name ) const
...
 {
     BOOST_FOREACH( const Target& target, targets )
     {
         const brion::Strings& values = target.get( name );
+        found = found || ( values.size() != 0 );
```
tribal-tec: !empty()

brion/target.cpp
```diff
@@ -171,7 +172,11 @@ void _parse( const Targets& targets, const std::string& name, GIDSet& gids )
 GIDSet Target::parse( const Targets& targets, const std::string& name )
 {
     brion::GIDSet gids;
-    _parse( targets, name, gids );
+    bool found(false);
+    _parse( targets, name, gids, found );
+    if( !found )
```
tribal-tec: isn't it easier to check for gids.empty() here which should lead to the same result? Why throw at all?

doc/Changelog.md
```diff
@@ -3,6 +3,14 @@ Changelog {#Changelog}
 
 # git master
 
+* [45](https://github.com/BlueBrain/Brion/pull/45):
+  Made targets optional. brain::Circuit::getGIDs() does not require targets
```
tribal-tec: Too long :) Usually the brief of the commit, if you want to write more, put in the commit message in a new paragraph.

brain/circuit.h
```diff
@@ -61,10 +61,8 @@ class Circuit : public boost::noncopyable
 
     /**
      * @return The set of GIDs for the given target name. If empty it will
```
tribal-tec: The brain::Circuit should not deal with targets. There is brion::Target class for that. Circuit::getGIDs() should return all GIDs of the circuit. Simple, easy, there you go!

hernando: I don't get what you mean by this. Are you saying that getGIDs() should take no arguments? If that's the case I don't agree with you at all. As a user I'd be very annoyed by having to deal with all the code for creating a blue config, then the target parsers based on the information provided by the blue config, and then call brion::Target::parse. If getGIDs() takes no arguments all the code in brain::Circuit that is already simplifying creeps from where it's nicely hidden into user code.

brion/blueConfig.cpp
```diff
@@ -332,8 +339,7 @@ std::string BlueConfig::getCircuitTarget() const
 
 GIDSet BlueConfig::parseTarget( const std::string& target ) const
 {
-    return brion::Target::parse(
-        getTargets(), target.empty() ? _impl->getCircuitTarget() : target );
+    return brion::Target::parse( getTargets(), target );
```
hernando: Did we agree on changing also this? This is affecting much more code than only Fivox, as the change in brain::Circuit does.

brion/target.cpp
hernando: Have you considered making _parse throw? Maybe the code is simpler, overall if both Target::parse and Target::get throw.

brion/target.h
```diff
@@ -67,6 +67,7 @@ class Target
      *
      * @param name target name to get the values from
      * @return list of values in the target. Can be targets and/or GIDSet
+     * @throw std::runtime_error if name is an invalid target
```
hernando: If this throws, Target::parse should also throw. The funny part is that you've implemented the throw in parse but not in get.

brain/circuit.h
tribal-tec: Final conclusion: two functions
- getGIDs(): returns all GIDs of the circuit
- getGIDs( target ): GIDs of given target, throw if target cannot be found (no parsers or invalid target name)

brion/blueConfig.cpp
tribal-tec: This change seems reasonable and we should fix according code respectively. If one wants to parse the circuit target, it should be explicitly called like that.

brion/target.cpp
tribal-tec: Probably the best solution. Dealing with the return value vs recursion is annoying probably

brion/blueConfig.cpp
hernando: Fine with me then.

favreau:
Ping

hernando:
You have to rebase.

hernando:
+1 otherwise

PR #47, opened 2016-03-11
Update MVDTool tag because current tag doesn't set Boost's inclusion …

ptoharia:
…directories.


eile:
Retest this please.

eile:
Seems to need include_directories(SYSTEM)

PR #51, opened 2016-03-18
Fix: Empty target is valid

favreau:


tribal-tec:
Squash all the 4 commits to one?

favreau:
Squashed!

PR #54, opened 2016-03-29
Release Brion-1.7.0

chevtche:


doc/Changelog.md
```diff
@@ -13,6 +15,14 @@ Changelog {#Changelog}
...
   - brain::Circuit::getNumNeurons() returns the total number of neurons in the
     circuit.
+* [43](https://github.com/BlueBrain/Brion/pull/43):
+  Add MVD3 support to Brain::circuit 
+  - Add mvd-tool as a subroject for MVD3 parsing
+  - Change brain::getTransforms logic from euler angles to rotation with quaternions
+  - Add switch in BlueConfig to load or not MVD3
+  - Add tests for MVD3 parsing
+  - Add Nix support
+  - Switch MVDTool dependency to github repository
```
eile: Too long. This is a Changelog, if I want details on a change I go to the PR. Just remove the bullet points.

brain is lowercase (namespace)

doc/Changelog.md
```diff
@@ -1,8 +1,10 @@
 Changelog {#Changelog}
 =========
 
-# git master
+# Release 1.7.0 (24-Mar-2016)
 
+* [51](https://github.com/BlueBrain/Brion/pull/51):
+  Empty target does not rise an exeption anymore
```
eile: Again, not helpful (as in the commit msg already). Think of what you want to know reading this, not what you did writing the code. Should be something 'Empty target does foo instead of raising an exception'

@favreau as the original committer.

doc/Changelog.md
chevtche: Ok. Changed, but some changes has bullet points

chevtche:
Updated

doc/Changelog.md
eile: that's too bad.

eile:
The contract for Target::parse is:
* throw if a non-existent target is found
* empty targets are valid, i.e., do not throw when an empty target is found
* empty target name will throw, since it cannot exist in a file (no reason to document this)

Note that this means that the outer try/catch probably needs to go.

@chevtche please document, test and implement this accordingly in a separate PR
@hernando ok with you?

eile:
Why closed?

PR #61, opened 2016-04-11
New vmmlib API

eile:


brain/neuron/soma.h
```diff
@@ -28,6 +28,7 @@
 
 namespace brain
 {
+/** High-level interface to neuron morphologies. */
```
tribal-tec: moved that to types.h? namespace brain brief is there as well

tribal-tec:
+1 otherwise

brain/neuron/soma.h
eile: why is there no brain/neuron/types.h.... sigh.

eile: done

PR #63, opened 2016-04-18
Move old SDK/Monsteer spike report here pending refactoring

eile:


bilgili:
What is the reason for this change ? it is not clear from the topic.

rdumusc:
If I remember correctly the Spikes class were never supposed to be part of Monsteer, they had just been moved out of BBPSDK to some "random" location (Monsteer) for open sourcing Fivox.
Now I was going to suggest some improvements to this API while we are moving it, then I realised there had been this long discussion about the new Spikes API in Brain here: https://github.com/BlueBrain/Brion/pull/44. So I am also not sure why we are moving this now.
A changelog entry should explain the purpose of this commit, including the fact that it is temporary, with a link to the new implementation proposal.

eile:
1) Because already we are forgetting its purpose and why its in Monsteer
2) To get the dependencies right (Fivox depended on Monsteer!?)
3) Its logical place is here, now that libBrain exists

All this is independent to #44, which is needed too. I would rather not changelog it at this point, since it just draws unjust attention to it.

bilgili:
- This was not part of the spring cleanup planning
- I dont see the necessity right now until we need it

PR #68, opened 2016-05-04
Specification document for new CompartmentReport.

hernando:


hernando:
@eile, @bilgili, @jafyvilla, please review.

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,422 @@
...
+         *
+         * Preconditions:
+         * - The report was open in write mode.
+         * - writeMapping has already been called.
+         *
+         * @param gid the GID of the cell
+         * @param voltages the voltages per compartment to save
+         * @throw std::runtime_error if writing was not successful.
+         */
+        void writeCell( const uint32_t gid, const floats& voltages );
```
rdumusc: Just write() ? Or rename writeMapping() to writeCellMapping() for consistency?

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,422 @@
...
+         * The neuron index is derived from the order in the GID set provided by
+         * updateMapping().
+         *
+         * @param index neuron index per current GID set
+         * @return number of compartments for the given neuron or 0 if the
+         *         report has just been created or updateMapping is called, and
+         *         not frame has been read yet. Also 0 for neurons not part of
+         *         the report at the source.
+         */
+        size_t getNumCompartments( const size_t index ) const;
```
rdumusc: Is this the same as getCompartmentCounts()[index]? If so is it bringing much? It's also the only function which takes an index instead of a gid, slightly surprising for users.

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,422 @@
...
+         * completed.
+         *
+         * This operation may throw away cached data and make getCurrentTime()
+         * jump a step bigger than the timestep for the next read. It is
+         * implementation defined when this may happen.
+         *
+         * @throw std::runtime_error is the report was open in write mode or
+         *        there's an on-going read operation.
+         */
+        void updateMapping( const GIDSet& gids );
```
rdumusc: Maybe we should find a different name which does not include mapping. This is confusing with the writeMapping() function below where mapping has a different semantics.

rdumusc:
I was not in the list of reviewers but I had a quick look. Just minor comments but the otherwise this API looks good.

doc/feature/compartmentReport.md
hernando: What about selectSubset?

doc/feature/compartmentReport.md
hernando: No, it's not the same. getCompartmentCounts()[index] gives a uint16_ts with the number of compartments on each of the sections of the neuron with that index. The index appears also in the two functions above, but not explicitly.

I wondered whether getCompartmentCounts() and getOffsets() should take a gid and not return a vector of vectors. This approach requires the implementation to store a translation table that may not be needed depending on how the user processes the simulation data. Since this is supposed to be the low level API, I decided to keep it as it is right now.

doc/feature/compartmentReport.md
hernando: I prefer writeCellMapping. Simply write suggests a sense of totality that is not accurate.

rdumusc: Good for me, I think that both work and in general I like longer explicit name. 
Here however write() reads write(cell_gid, voltages) in practice which I don't find too "totalitarian" :-)

doc/feature/compartmentReport.md
rdumusc: +1 sounds good!

doc/feature/compartmentReport.md
rdumusc: OK got it, this function returns the sum(getCompartmentCounts()[index]). I am not sure if its going to be needed, but why not - I'm ony slighty worried about the risk of mixing indices with gids when calling this function.

In the general case returning a vector of vectors looks ok, but as you say both approaches have pros and cons. It mostly depends on how it is going to be used - the idea is of course to minimize the amount of user code needed to access the data. If users are going to often retrieve data for a single GID, they have to do the mapping themselves and in the case the API should do it instead.

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,422 @@
...
+SpikeReport does. There is a network plugin implementated with codash, but it
+has been deprecated. Since there's a new specification of the SpikeReport API
+ready for implementation, it makes sense to review the CompartmentReport
+interface to make it consistent with the new SpikeReport and prepare it for the
+ZeroEQ restrictions at the same time.
+
+This specification proposes a new interface for the CompartmentReport with
+these goals in mind.
+
+## API
```
eile: Is it maybe easier to just list the changes to the API?

doc/feature/compartmentReport.md
eile: IMO only if we need to change this method anyways.

doc/feature/compartmentReport.md
hernando: It depends on what type of spec and review you want to do. If you want to use this proposal to make  an incremental update on the current API it's better to list the differences, but it will make the document possibly harder to read. The document as is is easier to read with no prior knowledge.

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,422 @@
...
+         * Postconditions:
+         * Let:
+         *  - r be the CompartmentReport
+         *  - f be the returned future by r.seek(toTimestamp)
+         *  After f.wait() returns:
+         *  - r.getCurrentTime() ~= toTimestamp rounded to the nearest multiple
+              of the timestep, ...
+         *  - ... getState() == OK -> r.getEndTime() > getCurrentTime() and ...
+         *  - ... getState() == ENDED -> r.getEndTime() == getCurrentTime()
+         *
```
jafyvilla: I guess that here you wanted to separate and explain all the different possibilities, but it took me some time to understand, maybe because of the format. Probably phrasing each case in a couple of sentences would help (only for future references and/or doxygen)

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,422 @@
...
+         *    empty
+         *  - Otherwise r.getCurrentTime() ~= s + r.getTimestep() (the timestamp
+         *    is rounded to the nearest multiple of the timesetp).
+         *  - The mapping data becomes defined.
+         *
+         * After successful f.wait_for or f.wait_until, the result is the same
+         * as above. If they time out, no observable state changes.
+         *
+         * If the state is FAILED or some other operation is still pending
+         * when read is called the result is undefined.
```
jafyvilla: It never throws? even if the report was not open in read mode?

doc/feature/compartmentReport.md
eile: To me, a spec describes a new feature. In the case of the SpikeReport that was a completely new API. In case of this, it feels more like a delta and the current form makes it hard to understand what changes. (eg http://eyescale.github.io/equalizergraphics.com/documents/design/roi.html)

If I have no prior knowledge, I can read the full API doc in the .h.

doc/feature/compartmentReport.md
hernando: For consistency with the SpikeReport it should throw logic_error if preconditions are not met instead of having an undefined result.

jafyvilla: That's why I asked, the throw is not documented here (it is in others)

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,422 @@
...
+
+        /** Write the compartment counts for each section for one cell.
+         *
+         * Preconditions:
+         * - The report was open in write mode.
+         * - writeMapping has already been called.
+         *
+         * @param gid the GID of the cell
+         * @param voltages the voltages per compartment to save
+         * @throw std::runtime_error if writing was not successful.
```
jafyvilla: For the write methods, wouldn't it be better to return the result of the operation as a bool? and keep the throw for when things go really wrong.

That way the user could do things like "if write() then foo" or "if !write() then bar"

hernando: Can you give an example of when you would like to really write that code. For me a write operation is something that either succeeds or fails, you can use either a return value or a exception to indicate a failure, but I don't find a good reason for using both.

jafyvilla: Mmm... I can't think of a specific example now. But in general, my idea is that exceptions are for really exceptional behaviors, as their impact is usually more critical than a return value, and their handling a bit more complicated.
In the case of read/write functions, as you said, they can either succeed or fail, and those are very valid results that usually happen often.

There's no reason to use both exceptions and return values, true (maybe only if, internally, there is something specific that exceptionally failed and you want to make sure that is handled, then the exception can be raised). But in my opinion, and without knowing about the best practices in this context, I guess I tend to prefer return values. That said, I have no strong preference, and since you have much more experience in the usage of this kind of apis, you are in a better position to decide :)

hernando:
I think I'll keep the exceptions. Using a return value forces the user to write code to check it because not doing it can lead to strange behaviour later on if an error occurred. If an exception is not handled it will make the program abort, but at least it won't produce incorrect results silently.

eile:
> I think I'll keep the exceptions.

Not sure what you are referring to, but (without context) I don't agree on the following conclusions. Not checking for error leads to undefined behaviour anyways (crash...) and writing try/catch is awkward for error handling imo. There is always teh ```__attribute__( don't ignore return value )``` if we want to enforce error handling.

hernando:
Stating "writing try/catch is awkward for error handling imo" is the awkward thing to me because exceptions (and try/catch for that matter) were invented exactly for error handling. An error to me is a condition that cannot be considered the normal flow of a program, in this sense error==exception, and using return values for errors leads to very verbose code.

Maybe you missed the context because I think I didn't replied to Jafet's last comment, but added a general comment instead.

doc/feature/compartmentReport.md
hernando: I think we have to discuss it in person. For the next time it may be better to do as you suggest, but changing the whole spec now to be a diff is going to be too much work for what I can see it's worth (it's going to be too much formatting work for no new content for a document that won't be touched after acceptance). You have to convince me that it will really pay off during the implementation.

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+## Implementation
+
+This section will cover some implementation details related to the stream
+case as the file based IO is straight forward.
+
+The most two important decisions taken are:
+- The granularity of the messages sent won't be the whole frame but a cell.
+- The mapping will be always included as part of the frame.
```
eile: Of each frame?

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,422 @@
...
+both in the client side and in the implementation of existing plugins.
+Therefore, the decision here is to keep the API similar to the current one
+and send one message per cell. This allows interesting things like filtering
+cell subsets on origin based on topics, something much more difficult to do
+with whole frames. It also allows very easily having multiple clients receiving
+different portions of the circuit and only those. Obviosly this introduces an
+overhead, but we expect it to be small in relative and absolute terms. At the
+same time we don't expect to stream full compartment data for detailed
+morphologies for more than 10^5 because for larger circuits volumetric
+approaches with voxelization at origin are more reasonable. Having a separate
```
eile: Note that "at origin" might mean "in different processes and nodes in the same supercomputer" due to memory pressure in the simulation.

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,422 @@
...
+
+The current API for writing is already cell oriented, while this is convenient
+to write on memory mapped files or H5 files without requiring the client to
+assemble the whole buffer before writing, it's not consistent with the fact that
+reads are whole frame oriented. This means that in the stream case we have to
+assemble the frame at the writer or reader part of the implementation, or
+change the API and force the client to do it on the writer side. Requiring
+the client to assemble the frame is the easiest solution for the implementation
+of the stream based case but it requires bigger changes to adapt existing code
+both in the client side and in the implementation of existing plugins.
```
eile: client side is confusing. If I understand correctly, the assembly happens at the receiver end, but within the plugin?

doc/feature/compartmentReport.md
eile: Yes :)

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,422 @@
...
+compartments per section and 4 compartments on average, the overhead is an
+additional 1/32th of the frame size. In this case, for every 4 float numbers
+(the compartment data) we have to add a compartment count that can be enconded
+with 4 bits, in other words we have 4/(4*32) extra bits for each data bit. The
+actual overhead still needs to be figured out, but anything around this 3%
+seems reasonable, overall considering that is data size is a problem a big
+reduction can be achived using compression of the floating point data.
+
+The messages of the protocol are:
+ - Publiser to subscribers:
```
eile: +h

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,422 @@
...
+   * FRAME_BARRIER
+   * REPORT_METADATA
+ - Subscribers to publisher:
+   * REQUEST_METADATA
+
+The fields of a frame data message are the following:
+ - cell GID
+ - timestamp
+ - compartment data (n floats)
+ - compartment counts (m integers packed as 4 or 8 bit n = Sum counts)
```
eile: Intuitively I would expect the last two in reverse order.

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,422 @@
...
+
+The timestamp is guaranteed to be the same for all cells in the same frame which
+means that a frame can be considered finished when the last received timestamp
+is different to the previous one. Nevertheless, in order to avoid receivers wait
+until the first message of the next frame arrives, the FRAME_BARRIER message
+is used to signal the end of a frame (this message is an empty event). The
+FRAME_BARRIER message is used before the very first frame data message, this
+is done because otherwise receivers do not have enough information to know
+if the very first frame is a complete frame or they joined late.
```
eile: FRAME_START with timestamp as payload then?

doc/feature/compartmentReport.md
hernando: Client refers to the user of the API.

doc/feature/compartmentReport.md
hernando: I don't know if I need to care about this, but I did it on purpose to avoid memory alignment issues if we decide to use a char array to encode (and possibly compress) the compartment counts.

doc/feature/compartmentReport.md
hernando: The timestamp is already part of the frame message on purpose (so if this message is missed nothing bad can happen). Adding it here is redundant. Removing it from the other message makes the protocol stateful.

doc/feature/compartmentReport.md
hernando: I don't get why it's important what "origin" could be. Can you please suggest a new wording to clarify.

doc/feature/compartmentReport.md
eile: It seems redundant, but may also be added robustness. In Collage I've seen weird cases where packets arrive late, so being able to correlate a barrier to a frame explicitly instead of implicitly is imo a good idea.

hernando:
Welcome back @tribal-tec! You're kindly invited to take a look at the spec and make comments. I've restructured it as a diff as @eile requested, it should be easy to read top bottom.

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,467 @@
...
+       boost::future< void > seek( const float toTimestamp );
+
+       /** Finish writing the current frame and advance to the next one.
+        *
+        * Upon return getCurrentTime() has been increased to the nearest
+        * multiple of the timestep greater than the value before.
+        */
+       void endFrame();
+
+- Replace loadFrame(float) by loadNextFrame():
```
tribal-tec: function is called readNextFrame()

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,467 @@
...
+        *
+        * After successful f.wait_for or f.wait_until, the result is the same
+        * as above. If they time out, no observable state changes.
+        *
+        * @throw std::logic_error if a precondition is not met..
+        * @version TBD
+        */
+       boost::future< void > seek( const float toTimestamp );
+
+       /** Finish writing the current frame and advance to the next one.
```
tribal-tec: frame vs timestamp (again). seek() uses timestamp, end/readNext frames. Looks inconsistent at first, maybe some clarification about the concepts is necessary. Should the terms also be reflected in the function names?

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,467 @@
...
+        * - The report was open in write mode.
+        * - writeHeader has already been called.
+        *
+        * @param gid the GID of the cell
+        * @param counts the number of compartments per section
+        * @throw std::logic_error if a precondition is not met and
+        *        std::runtime_eror if an IO error occurs.
+        * @version TBD
+        */
+       void writeCellMapping( const uint32_t gid,
```
tribal-tec: It updates the internal mapping, but with this call only it's more like writeCellComparments() or writeCellCompartmentCounts(). I'm not sure the term 'mapping' is helpful for a user with our backgrounds.

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,467 @@
...
+        *  simulation frame buffer.
+        *
+        * For instance, getOffsets()[1][15] retrieves the lookup index for the
+        * frame buffer for section 15 of neuron with index 1. The neuron index
+        * is derived from the order in the GID set provided by updateMapping().
+        *
+        * After construction or updateMapping is called, the result is empty
+        * until a frame is read.
+        *
+        * For requested ceels that are not part of the report its section
```
tribal-tec: typo ceels

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,467 @@
...
+This section will cover some implementation details related to the stream
+case as the file based IO is straight forward.
+
+The most two important decisions taken are:
+- The granularity of the messages sent won't be the whole frame but a cell.
+- The mapping will be always included as part of the frame.
+
+The rationale for these decisions follows.
+
+The current API for writing is already cell oriented, while this is convenient
```
tribal-tec: I'm wondering if  you have discussed the possibility to make the API wrt read & write symmetric, meaning read also works cell oriented, not frame oriented. From my short-after-vacation point of view it could eliminate the need for updateMapping(), plus also potential allocations, buffering, aggregation for received data. The end-user most likely wants to have a full frame for a set of cells, but this can be achieved again through a higher-level API. For the low-level read/write, cell-based seems more straightforward and easier to implement, understand and maintain.

hernando: It came to my mind, but it hasn't been discussed. My feeling is that some buffering will still be needed and the synchronization at frame boundaries may be more difficult if the API doesn't have a function to read a full frame atomically. If you want to have a whiteboard discussion I'm open to it, otherwise feel free to provide a more concrete proposal.
Something that was proposed and probably is a good idea is to make all mapping functions get a GID as input parameter. That frees the user from having to do the GID->index translation, which has already caused confusion in the past.

doc/feature/compartmentReport.md
hernando: The class description has been extended. Let me know if that's sufficient.
Note that the API functions that refer to frames do not use frame numbers at all. This means that how timestamps are discretized into frames is kept inside the implementation.

tribal-tec: I saw the class description later, it's sufficient then.

doc/feature/compartmentReport.md
tribal-tec: I also prefer renaming.

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,470 @@
...
+        *  - ... getState() == OK -> r.getEndTime() > getCurrentTime() and ...
+        *  - ... getState() == ENDED -> r.getEndTime() == getCurrentTime()
+        *
+        * After successful f.wait_for or f.wait_until, the result is the same
+        * as above. If they time out, no observable state changes.
+        *
+        * @throw std::logic_error if a precondition is not met..
+        * @version TBD
+        */
+       boost::future< void > seek( const float toTimestamp );
```
tribal-tec: forward()

doc/feature/compartmentReport.md
```diff
@@ -0,0 +1,470 @@
...
+        *
+        * After successful f.wait_for or f.wait_until, the result is the same
+        * as above. If they time out, no observable state changes.
+        *
+        * @throw std::logic_error if a precondition is not met.
+        * @version TBD
+        */
+       boost::future< floatsPtr > readNextFrame( );
+
+- Rename updateMapping() to selectCellSet and rewrite it's documentation to be:
```
tribal-tec: remove dynamic cell selection, only via ctor (same as SpikeReport).

hernando:
Updated.

doc/feature/spikeReport.md
```diff
@@ -185,26 +185,31 @@ Thread-safey is only guaranteed for const access.
...
-         *    it at all if possible).
+         *  - The postconditions of read operations imply that in forward skips
+         *    this function throws away the data previous to toTimestamp (or
+         *    avoids reading it at all if possible).
          *
-         * @throw std::runtime_error if a precondition does not hold.
+         * @throw std::runtime_error if a precondition does not hold or the
+         *        operation is not supported by the implementation.
          */
         boost::future< void > forward( float toTimestamp );
```
tribal-tec: rename to seek()

hernando:
Ready to merge?

tribal-tec:
For me yes.

PR #70, opened 2016-06-09
Fix MSVC warning C4373 (virtual function overrides).

ptoharia:


ptoharia:
Servus' pull request needed to remove some MSVC errors. Still some other errors.

brion/plugin/compartmentReportBinary.cpp
```diff
@@ -204,7 +204,7 @@ size_t CompartmentReportBinary::getFrameSize() const
     return _subtarget ? _comps : _header.numCompartments;
 }
 
-floatsPtr CompartmentReportBinary::loadFrame( const float timestamp ) const
```
eile: Should stay const here. Did you get a warning? If yes, which one?

ptoharia: I was getting the virtual override and I changed all the prototypes, both in .h and .cpp files. I guess I should have changed only the headers, right? I'll roll back cpp changes if so

eile: Yep. I don't have the link at hand, but basically the header is for users (const on by-value is pointless) and the impl is for developers (const means the copy param is not modified)

ptoharia:
@eile rolled back cpps

eile:
Retest this please.

eile:
+1

PR #72, opened 2016-06-13
Morphology version 1.1 specification

tribal-tec:


doc/feature/morphology11.md
```diff
@@ -0,0 +1,128 @@
...
+        FAMILY_GLIA = 1
+    };
+
+    class Morphology
+    {
+        ...
+        /**
+         * @return the cell family of that morphology. FAMILY_GLIA only for
+         *         MORPHOLOGY_VERSION_H5_1_1 with a perimeter dataset,
+         *         otherwise FAMILY_NEURON.
```
hernando: Not clear. I'd simply say "the cell family of this morphology". The rest is an implementation detail and I as I read is not fully precise. Neurons can also have a perimeter according to format 1.1, but it's optional in this case. On the other hand, the dataset is mandatory for glial cells.

doc/feature/morphology11.md
```diff
@@ -0,0 +1,128 @@
...
+         */
+        CellFamily brion::getCellFamily() const;
+
+        /**
+         * @return perimeters of the cross sections for each point of the
+         *         morphology in micrometers. Valid only for cell family
+         *         FAMILY_GLIA.
```
hernando: I'm not sure this is necessarily true. The file format doesn't prevent the existence of the perimeter data set for neurons. As I said above, the only restriction documented in the spec is that it must exists for glial cells.

doc/feature/morphology11.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Issues
+
+### 1: Do we choose cell family automatically or do we require an explicit selection?
+
+_Resolved: Yes_
+
+To maintain the existing API, the cell family will be chosen automatically. The
+family will be FAMILY_GLIA after writePerimeters() or the value of the
+cell_family enum, or FAMILY_NEURON in all other cases.
```
hernando: You can add a new function and keep the semantics intact for old code. You just have to make FAMILY_NEURON the default. I prefer explicit rather than implicit in this case.

doc/feature/morphology11.md
```diff
@@ -0,0 +1,128 @@
...
+        floatsPtr readPerimeters() const;
+
+        /**
+         * Write perimeters of morphology. The cell family will be set to
+         * FAMILY_GLIA.
+         *
+         * @param perimeters perimeters of the cross sections for each point of
+         *                   the morphology in micrometers
+         */
+        void writePerimeters( const floats& perimeters );
```
eile: metadata API?

tribal-tec:
Updated. New constructor for writing which enforces CellFamily. Add an issue for metadata API.

doc/feature/morphology11.md
```diff
@@ -0,0 +1,138 @@
...
+
+_Resolved: Yes_
+
+As the perimeters can also be optionally used for the neuron cell family, a new
+constructor for writing morphologies enforces to explicitly choose the cell
+family.
+
+### 2. Why is the metadata not exposed in the API?
+
+_Resolved: Yes_
```
eile: For future reference: Resolved was ambigious. Hello now has 'Resolution: $WHATISTHERESOLUTION. $LONGEXPLANATION.'

eile:
+1 from me.

tribal-tec:
Adapted the examples, fixed the 'Resolved' parts.

doc/feature/morphology11.md
```diff
@@ -0,0 +1,139 @@
...
+         * @param family the cell family that this morphology represents.
+         */
+        Morphology( const std::string& file, CellFamily family );
+
+        /** @return the cell family of that morphology. */
+        CellFamily brion::getCellFamily() const;
+
+        /**
+         * @return perimeters of the cross sections for each point of the
+         *         morphology in micrometers.
```
hernando: What happens if there aren't any? Throw or empty result?

tribal-tec:
Added throw for read constructor if FAMILY_GLIA and no perimeter dataset

doc/feature/morphology11.md
```diff
@@ -0,0 +1,142 @@
...
+
+    class Morphology
+        // Default parameter brion::MorphologyRepairStage for all functions to
+        // brion::MORPHOLOGY_UNDEFINED. Needed only for V2 and throws there if
+        // brion::MORPHOLOGY_UNDEFINED.
+
+    /**
+     * ...
+     * @throw std::runtime_error if morphology if of family FAMILY_GLIA and
+     *                           does not contain perimeters dataset
```
hernando: typo if->is.
I would simply say that this function may throw if the file contains errors (you can then give the missing perimeters as an example).

tribal-tec: Then the current throw description still applies: "throw std::runtime_error if file is not a valid morphology file"

tribal-tec:
Existing throw documentation for read-constructor applies also for glia morphology w/o perimeter dataset. No API change required, implementation throws accordingly.

doc/feature/morphology11.md
hernando: I think you should still add here that this function may return an empty vector if CellFamily is neuron.

PR #75, opened 2016-06-15
Implement morphology version 1.1 specification

tribal-tec:


brion/morphology.h
```diff
@@ -81,6 +86,14 @@ class Morphology : public boost::noncopyable
      */
     BRION_API Vector2isPtr readApicals() const;
 
+    /**
+     * @return perimeters of the cross sections for each point of the morphology
+     *         in micrometers.
+     * @throw std::runtime_error if empty for FAMILY_GLIA
```
hernando: Didn't we agree that the existence of the dataset should be checked at construction?
But this is also fine, because I don't think we are checking the consistency of any file in the constructor.

tribal-tec: Hmm, good point. Ideally we should check what we can check in the constructor. Will add it, at least for version 1 and 1.1.

brion/morphology.h
```diff
@@ -132,11 +156,22 @@ class Morphology : public boost::noncopyable
      *
      * @param apicals section and point index of all apical points
      * @throw std::runtime_error if object not created with write ctor
-     * @throw std::runtime_error if called for version 1 files
+     * @throw std::runtime_error if not supported by implementation
```
hernando: You can consider deprecating this function, it only makes sense for v2 morphologies as the other file formats don't store these points.

brion/morphology.h
```diff
@@ -95,10 +108,21 @@ class Morphology : public boost::noncopyable
      * @param overwrite true to allow overwrite of existing file
      * @throw std::runtime_error if file could not be created
      * @version 1.0
+     * @deprecated
      */
     BRION_API Morphology( const std::string& target,
                           MorphologyVersion version, bool overwrite = false );
```
hernando: I've realized that if we remove this constructor, we won't be able to write SWC morphologies in the future. Not that it really matters, as the writing functions in the SWC plugin are unimplemented, but just to make you aware of the fact.
We can keep the function deprecated to discourage usage until we have a clearer idea.

brion/morphologyPlugin.h
```diff
@@ -47,15 +48,29 @@ class MorphologyInitData : public PluginInitData
...
         return _version;
     }
 
+    CellFamily getFamily() const
+    {
+        return _family;
+    }
+
 protected:
     MorphologyVersion _version;
```
hernando: These attributes can be const, can't they?

brion/morphology.h
tribal-tec: Is it commonly agreed that v2 is deprecated? If not, this has to stay like this imo.

brion/plugin/morphologyHDF5.cpp
```diff
@@ -352,6 +372,41 @@ Vector2isPtr MorphologyHDF5::readApicals() const
...
+        floatsPtr data( new floats( dims ));
+        dataset.read( data->data(), H5::PredType::NATIVE_FLOAT );
+        return data;
+    }
+    catch( ... )
+    {
+        if( getCellFamily() == FAMILY_GLIA )
+            LBTHROW( std::runtime_error( "No empty perimeters allowed for glia "
+                                         "morphology" ));
+        return floatsPtr( new floats( ));
```
hernando: This line shouldn't be needed to avoid a warning if LBTHROW is declared as throwing. I don't know how to do that but _abort_ is doing the same kind of thing.

brion/plugin/morphologyHDF5.cpp
```diff
@@ -392,34 +447,28 @@ void MorphologyHDF5::writeSections( const Vector2is& sections,
     H5::FloatType structureDT( H5::PredType::NATIVE_INT );
     structureDT.setOrder( H5T_ORDER_LE );
```
hernando: You can move these two declarations just before line 456 (and do the same in other functions).

tests/morphology.cpp
```diff
@@ -180,12 +196,79 @@ BOOST_AUTO_TEST_CASE( h5_write_v1 )
...
+    path /= "local/morphologies/01.07.08/h5/R-C010306G.h5";
+
+    const brion::Morphology morphology( path.string( ));
+
+    const brion::floats perimeters;
+    brion::Morphology neuron( file, brion::FAMILY_NEURON );
+    neuron.writePoints( *morphology.readPoints( ));
+    neuron.writeSections( *morphology.readSections( ));
+    neuron.writeSectionTypes( *morphology.readSectionTypes( ));
+    neuron.writePerimeters( perimeters );
```
hernando: The implementation shouldn't allow to write a file with n points and 0 perimeters. If you write points or perimeters and then you write the other dataset, they should be consistent. To leave it empty simply don't write it.

brion/morphology.h
hernando: But if you deprecate the constructor to open a v2 morphology for writing, then you are effectively deprecating this function also.

brion/plugin/morphologyHDF5.cpp
```diff
@@ -492,18 +536,52 @@ void MorphologyHDF5::flush()
...
 
+void MorphologyHDF5::_resolveV1()
+{
+    _points = _file.openDataSet( "/" + _d_points );
+    const H5::DataSpace pointsSpace = _points.getSpace();
+    if( pointsSpace.getSimpleExtentNdims() != 2 ||
+        pointsSpace.getSimpleExtentDims( _pointsDims ) < 0 ||
+        _pointsDims[1] != _pointColumns )
+    {
+        LBTHROW( std::runtime_error( "Reading morphology file '" +
```
hernando: Opening instead of reading?

tests/morphology.cpp
hernando: You've just remove the writePerimeters call from the test, but that doesn't address my comment about the consistency. It's also true that you could argue that the other functions are not checking the consistency either.

tribal-tec:
Updated

hernando:
+1

PR #77, opened 2016-06-17
Add brain::Circuit::getMorphologyTypes() and brain::Circuit::getElectrophysiologyTypes()

tribal-tec:


hernando:
I know you don't like long names, but MTypes and ETypes aren't descriptive names, and I don't think that nomenclature is even universal. I'd prefer getMorphTypes and getElectroTypes.
Apart from that, what about adding functions to get the full list of morphological and electrophysiological types, otherwise this is like half the story.

brain/circuit.cpp
```diff
@@ -124,6 +126,8 @@ class Circuit::Impl
     }
 
     virtual Vector3fs getPositions( const GIDSet& gids ) const = 0;
+    virtual size_ts getMTypes( const GIDSet& gids ) const = 0;
+    virtual size_ts getETypes( const GIDSet& gids ) const = 0;
```
hernando: You can keep this names in the implementation if you want.

tribal-tec:
Updated

hernando:
+1

PR #81, opened 2016-07-06
Fix GID out-of-bounds handling for MVD3 in brain::Circuit

tribal-tec:


brain/circuit.cpp
```diff
@@ -288,18 +288,32 @@ class MVD3 : public Circuit::Impl
...
-        return results;
+        try
+        {
+            const ::MVD3::Positions& positions = _circuit.getPositions( range );
+            assign( range, gids, positions, results, toVector3f );
+            return results;
+        }
+        catch( const HighFive::Exception& )
+        {
+            LBTHROW( std::runtime_error( "Error in getPositions()" ));
```
hernando: The error message is not very specific. Does HighFive provide any error message in the exception?

tribal-tec:
Updated

PR #82, opened 2016-07-07
Improve performance when reading synapse attributes from non-merged files

tribal-tec:


brion/synapse.cpp
```diff
@@ -338,47 +350,73 @@ class Synapse : public boost::noncopyable
...
+                                            regx, subs );
+
+            const std::string basename =
+                       firstFile.substr( 0, firstFile.find_last_of( '.' ) + 1 );
+            for( boost::sregex_token_iterator j; i != j; )
+            {
+                const std::string& fileNumber = *i++;
+                const uint32_t fileGID = boost::lexical_cast< uint32_t >(*i++);
+                _fileMap[ fileGID ] = basename + fileNumber;
+            }
```
hernando: I'd move the parsing of the merge script to a separate function and call it from the constructor to cache the result because the filesystem is already accessed there to find the candidate files.

brion/synapse.cpp
```diff
@@ -314,13 +304,77 @@ class Synapse : public boost::noncopyable
...
+
+        const boost::regex regx( "\\$CMD -i \\$H5.(?<number>[0-9]+) -o "
+                                 "\\$H5 -s /a(?<gid>[0-9]+)" );
+        const int subs[] = {1, 2};
+        boost::sregex_token_iterator i( content.begin(), content.end(), regx,
+                                        subs );
+
+        const std::string basename =
+                         filename.substr( 0, filename.find_last_of( '.' ) + 1 );
+        for( boost::sregex_token_iterator j; i != j; )
```
hernando: rename j to end?

PR #83, opened 2016-07-21
Add brain::Circuit::getRandomGIDs()

tribal-tec:


brain/circuit.cpp
```diff
@@ -147,6 +150,26 @@ class Circuit::Impl
         return brion::Target::parse( _targetParsers, target );
     }
 
+    GIDSet getRandomGIDs( const float fraction,
+                          const std::string& target ) const
+    {
```
hernando: You must provide a way to make the result deterministic. I'd suggest to either add an environment variable to choose the seed or always use the same seed. In particular, I think the C++11 code is using the most random device possible by default.

brain/circuit.h
```diff
@@ -61,15 +61,24 @@ class Circuit : public boost::noncopyable
...
 
+    /**
+     * @return A random fraction of GIDs from the given target name.
+     * @throw std::runtime_error if the target cannot be found.
+     */
+    BRAIN_API GIDSet getRandomGIDs( float fraction,
+                                    const std::string& target ) const;
+
+    /** @return A random fraction of GIDs from the circuit. */
+    BRAIN_API GIDSet getRandomGIDs( float fraction ) const;
```
hernando: In the implementation you're reading the whole target first. Why not make this functions free functions then:

    randomSampling(circuit.getGIDs(), 0.1)

tribal-tec:
Updated

PR #84, opened 2016-08-08
Brain synapses specification

tribal-tec:


doc/feature/synapses.md
```diff
@@ -0,0 +1,189 @@
...
+
+class Circuit
+{
+    SynapsesStream getAfferentSynapses( const GIDSet& gids ) const;
+    SynapsesStream getEfferentSynapses( const GIDSet& gids ) const;
+    SynapsesStream getSharedSynapses( const GIDSet& preGIDs,
+                                      const GIDSet& postGIDs ) const;
+
+    // convenience for upper getXXX( gids ).read( gids.size( )).get()
+    Synapses getAllAfferentSynapses( const GIDSet& gids ) const;
```
hernando: I'd call it getAfferentSynapses and rename the other one above, which I'd consider less common, to getAfferentSynapseStream

doc/feature/synapses.md
```diff
@@ -0,0 +1,189 @@
...
+    float postDistance() const;
+    Vector3f postSurfacePosition() const;
+    Vector3f postCenterPosition() const;
+
+    float conductance() const;
+    float utilization() const;
+    float depression() const;
+    float facilitation() const;
+    float decay() const;
+    int efficacy() const;
```
hernando: Synapse GID is missing.

doc/feature/synapses.md
```diff
@@ -0,0 +1,189 @@
...
+};
+
+// sequentially & forwarding iteration through synapses for any number of GIDs
+class SynapsesStream
+{
+    bool eos() const;
+
+    // async load of synapses for count cells
+    // throws if current + count leads to eos
+    std::future< Synapses > read( size_t count = 1 ) const;
```
hernando: Lots of question here.

Is there any reason for throwing?
Any guarantees about the order in which synapses are read in relation to the iteration order of the GIDSet for the afferent and efferent?
In the case of the shared synapses, which is the maximum count, pre.size() * post.size()? And the iteration order? Could it happen that after calling read the returned container is empty?

doc/feature/synapses.md
```diff
@@ -0,0 +1,189 @@
...
+
+    // union
+    Synapses operator+( const Synapses& rhs ) const;
+    Synapses operator|( const Synapses& rhs ) const;
+
+    // disjunction (XOR)
+    Synapses operator^( const Synapses& rhs ) const;
+
+    // difference
+    Synapses operator-( const Synapses& rhs ) const;
```
hernando: What are the complexity guarantees of the set operations? Are they creating a copy?
What about returning a SynapseStream also from here, the non explicit constructor of Synapses would do the rest.

doc/feature/synapses.md
```diff
@@ -0,0 +1,189 @@
...
+    float utilization() const;
+    float depression() const;
+    float facilitation() const;
+    float decay() const;
+    int efficacy() const;
+};
+
+// read-only, moveable, non-copyable, contains all data of all synapses
+class Synapses
+{
```
hernando: Why not adding a conversion constructor (i.e. without explicit) for this class that takes a SynapseStream as input. Then the getAllXXX methods are not needed and one could write:

     Synapses s = circuit.getAfferentSynapses(cells)
     SynapseStream s = circuit.getAfferentSynapses(cells);

doc/feature/synapses.md
```diff
@@ -0,0 +1,189 @@
...
+    SynapsesStream getAfferentSynapses( const GIDSet& gids ) const;
+    SynapsesStream getEfferentSynapses( const GIDSet& gids ) const;
+    SynapsesStream getSharedSynapses( const GIDSet& preGIDs,
+                                      const GIDSet& postGIDs ) const;
+
+    // convenience for upper getXXX( gids ).read( gids.size( )).get()
+    Synapses getAllAfferentSynapses( const GIDSet& gids ) const;
+    Synapses getAllEfferentSynapses( const GIDSet& gids ) const;
+    Synapses getAllSharedSynapses( const GIDSet& preGIDs,
+                                   const GIDSet& postGIDs ) const;
```
hernando: It's not clear which number has to be passed to getSharedSynapses(pre, post).read( xxx ).get() to get the same effect than this function.

doc/feature/synapses.md
```diff
@@ -0,0 +1,189 @@
...
+const brain::Synapses& synapses =
+    circuit.getAllAfferentSynapses( circuit.getGIDs( "Layer1" ));
+for( const auto& synapse : synapses )
+    std::cout << synapse.postGID() << " " << synapse.decay() << std::endl;
+```
+
+## Implementation
+
+* Data owned by container in the form of the brion::SynapseMatrix provided by
+  Brion, loaded upon construction
```
hernando: SynapseMatrix is a per (afferent or efferent) cell container, that means that you can't simply store it without possibly incurring in a big overhead for the shared synapses case.

doc/feature/synapses.md
```diff
@@ -0,0 +1,189 @@
...
+## API
+
+```cpp
+
+class Circuit
+{
+    SynapsesStream getAfferentSynapses( const GIDSet& gids ) const;
+    SynapsesStream getEfferentSynapses( const GIDSet& gids ) const;
+    SynapsesStream getSharedSynapses( const GIDSet& preGIDs,
+                                      const GIDSet& postGIDs ) const;
```
hernando: I guess this functions always load the data and don't cache anything internally, so if I want to do some operations with targets I should keep the containers myself, right?
This also means that if I want to have quick access to the afferent and efferent sets of each cell involved in a projection (like BBPSDK provides), I'd have to call getSharedSynapses pre.size() + post.size() times
(first each pre cell against the full post set for the afferent synapses, and then the reverse for the efferent), which is not a big issues, but then, much more importantly, there's no way the implementation can avoid having each synapse twice in memory.

doc/feature/synapses.md
tribal-tec: That's most certainly an elegant solution, thanks!

doc/feature/synapses.md
tribal-tec: from the nrn_extra file you mean?

hernando: Yes, or as done in BBPSDK if not present.

doc/feature/synapses.md
tribal-tec: * w/o throw, there requested read is not complete. Could be an option as well, haven't much thought about both possibilities.
* No order guarantees implied
* Shared synapses need some more thinking, I have no concrete implementation even in mind
* The container can be empty, yes

doc/feature/synapses.md
tribal-tec: While getting a first implementation for the synapse stream using position arrays, storing as matrices is not feasible. The data will be stored the same way like the BBPSDK Synapse_Dataset.

doc/feature/synapses.md
tribal-tec: Done

jafyvilla:
+1 for me

doc/feature/synapses.md
```diff
@@ -0,0 +1,215 @@
...
+    float conductance() const;
+    float utilization() const;
+    float depression() const;
+    float facilitation() const;
+    float decay() const;
+    int efficacy() const;
+};
+
+// read-only, moveable, non-copyable
+class Synapses
```
hernando: Just a side note, I think it's implied by being const, but this class must be thread-safe internally in my opinion. Otherwise it's unusable with parallel algorithms.

doc/feature/synapses.md
```diff
@@ -0,0 +1,215 @@
...
+## Examples
+
+### Stream synapse positions for pathway density voxelization
+
+```cpp
+const brain::Circuit circuit;
+const brain::SynapsesStream& stream =
+   circuit.getProjectedSynapses( circuit.getGIDs( "mtypetarget_1" ),
+                                 circuit.getGIDs( "mtypetarget_2" ));
+std::future< brain::Synapses > future = stream.read();
```
hernando: I still would like to know how can I know which is the largest value I can pass to read in the stream returned by getProjectedSynapses.

doc/feature/synapses.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Implementation
+
+* Data owned by container in the form of the arrays (transformed from
+  SynapseMatrix), loaded upon construction
+* Synapse object stores index to get return data from (parenting) container
+* SynapsesStream::read() uses async read of all synapse data of the given
+  fraction of GIDs
+* OPT: getProjectedSynapses() loads the synapses from the smaller GIDSet and
+  uses the correct side (afferent/efferent) accordingly
```
hernando: Add dots to finish the sentences. I thought the first four lines where a paragraph in the beginning.

doc/feature/synapses.md
```diff
@@ -0,0 +1,215 @@
...
+## API
+
+```cpp
+
+namespace brain
+{
+    // < postGID, afferent_index >
+    typedef std::pair< uint32_t, size > SynapseGID;
+
+    enum SynapsesPrefetch
```
hernando: You haven't explained anywhere what is this used for.

doc/feature/synapses.md
tribal-tec: Done

doc/feature/synapses.md
```diff
@@ -0,0 +1,232 @@
...
+     * - The GID of the post-synaptic cell.
+     * - The index of the synapse in the array of afferent contacts
+     *   of the post-synaptic cell before pruning/filtering.
+     * GIDs are invariant regardless of how the structural touches are
+     * converted into functional synapses during circuit building.
+     */
+    typedef std::pair< uint32_t, size > SynapseGID;
+
+    // loading of data during SynapseStream.read(), otherwise load happens on
+    // first touch
```
hernando: pre and post GIDs are always loaded by SynapseStream.read().

This needs to be more specific about what is a position attribute (because yes, they are also attributes), for example, postSectionGID is an attribute or a position (it's a topological position)?

tribal-tec:
Updated. Good to go?

doc/feature/synapses.md
```diff
@@ -0,0 +1,240 @@
...
+    float conductance() const;
+    float utilization() const;
+    float depression() const;
+    float facilitation() const;
+    float decay() const;
+    int efficacy() const;
+};
+
+// read-only, moveable, non-copyable, thread-safe
+class Synapses
```
hernando: Since it's read-only maybe it could be made copyable.
How is the moveable property implemented, C++11?

doc/feature/synapses.md
```diff
@@ -0,0 +1,240 @@
...
+BOOST_CHECK_EQUAL( synapses.size(), 5 );
+```
+
+## Implementation
+
+* Data owned by container in the form of the arrays (transformed from
+  SynapseMatrix), loaded upon construction wrt prefetch, lazy on first touch
+  except for pre- and postGID for connectivity queries and to fulfill the
+  contract for size().
+* Synapse object stores index to get data from (parenting) container.
```
hernando: Don't forget in the implementation that if a Synapse object touches the dataset it has also to perform the lazy loading. Maybe it's worth commenting it here.

doc/feature/synapses.md
tribal-tec: Yes, C+11 move.

PR #85, opened 2016-08-17
Implement brain synapses specification

tribal-tec:


README.md
```diff
@@ -45,15 +45,17 @@ and writing files which store the Blue Brain data model.
...
 The higher level library is called Brain and it provides:
 
-* brain::Circuit to facilitate loading information about cells and
-  morphologies in local and global circuit coordinates.
-* brain::neuron::Morphology with higher level functions to deal with morphologies.
+* brain::Circuit to facilitate loading information about cells, morphologies (in
+  local and global circuit coordinates) and synapses.
+* brain::neuron::Morphology with higher level functions to deal with
+  morphologies.
+* brain::Synapses and brain::Synapse for object and array access to synapses.
```
hernando: array and object

brain/circuit.cpp
```diff
@@ -662,14 +679,78 @@ Matrix4fs Circuit::getTransforms( const GIDSet& gids ) const
...
+{
+    const size_t i = afferent ? 0 : 1;
+    if( !_impl->synapseAttributes[i] )
+        _impl->synapseAttributes[i].reset( new brion::Synapse(
+                                               _impl->_synapseSource.getPath() +
+                             (afferent ? afferentFilename : efferentFilename)));
+    return *_impl->synapseAttributes[i];
+}
+
+const brion::Synapse* Circuit::getSynapseExtra() const
```
hernando: Throw instead of return 0? It looks a bit asymmetrical otherwise.

hernando: Now I see it's private, forget about it.

brain/circuit.h
```diff
@@ -120,15 +120,55 @@ class Circuit : public boost::noncopyable
...
     class Impl; //!< @internal, public for inheritance MVD2/3 impls
 
 private:
-    Impl* _impl;
+    friend class Synapses;
+
+    const brion::SynapseSummary& getSynapseSummary() const;
+    const brion::Synapse& getSynapseAttributes( bool afferent ) const;
+    const brion::Synapse* getSynapseExtra() const;
+    const brion::Synapse& getSynapsePositions( bool afferent ) const;
```
hernando: Move these functions to the Impl object, that will save a lot of _impl-> text in the implementation and you can change the implementation later without touching this file.

brain/synapse.h
```diff
@@ -0,0 +1,138 @@
...
+{
+public:
+    /** Copy construct this synapse by sharing the underlying data. */
+    BRAIN_API Synapse( const Synapse& );
+
+    /** Move construct this synapse. */
+    BRAIN_API Synapse( Synapse&& rhs );
+
+    /** Move assign this synapse. */
+    BRAIN_API Synapse& operator=( Synapse&& rhs );
```
hernando: Why no normal assignment?

brain/synapse.cpp
```diff
@@ -0,0 +1,175 @@
...
+{
+    return _impl->_synapses.postDistance()[_impl->_index];
+}
+
+Vector3f Synapse::getPostsynapticSurfacePosition() const
+{
+    return Vector3f( _impl->_synapses.postSurfacePositionX()[_impl->_index],
+                     _impl->_synapses.postSurfacePositionY()[_impl->_index],
+                     _impl->_synapses.postSurfacePositionZ()[_impl->_index] );
+}
```
hernando: In BBPSDK the positions where computed if the data file was not available. Should we do the same or can we safely assume that all circuits have position files (that was not true before)?
In some cases it might be cheaper to compute the position rather than loading it, but that requires loading the morphology (both pre and post for surface positions) and it only makes sense if the pre post section, segment ids are needed for something else, otherwise it's actually better to simply load the position.
Maybe the best solution is to provide some free functions that are able to compute the positions if needed.

brain/synapse.h
```diff
@@ -0,0 +1,138 @@
...
+    //@}
+
+private:
+    Synapse& operator=( const Synapse& ) = delete;
+
+    friend class Synapses;
+    Synapse( const Synapses& synapses, size_t index );
+
+    struct Impl;
+    std::unique_ptr< const Impl > _impl;
```
hernando: I'm hesitating whether you should use pimpl here. I don't expect many changes in the implementation of this object and for performance reasons it's better to not use the heap and simply store the Synapses& and index here. On the other hand, if performance is a concern, you should use the array interface in Synapses instead.

brain/synapse.cpp
```diff
@@ -0,0 +1,175 @@
...
+{
+
+struct Synapse::Impl
+{
+    Impl( const Synapses& synapses, const size_t index )
+        : _synapses( synapses )
+        , _index( index )
+    {}
+
+    const Synapses& _synapses;
```
hernando: This is a bit unconvenient for the wrapping, I'll have to think about how to deal with memory management. Since probably I'll have to introduce some type of shared pointer I think it's preferable to not use pimpl here to reduce the amount of heap objects that are needed.

brain/synapse.h
```diff
@@ -0,0 +1,138 @@
...
+    BRAIN_API Synapse( const Synapse& );
+
+    /** Move construct this synapse. */
+    BRAIN_API Synapse( Synapse&& rhs );
+
+    /** Move assign this synapse. */
+    BRAIN_API Synapse& operator=( Synapse&& rhs );
+
+    BRAIN_API ~Synapse();
+
```
hernando: What about an operator==. Note that comparisons between synapses from different containers are reasonable as long as they come from the same circuit.

brain/types.h
```diff
@@ -56,5 +60,32 @@ typedef std::vector< Quaternionf > Quaternionfs;
...
+ * GIDs are invariant regardless of how the structural touches are
+ * converted into functional synapses during circuit building.
+ */
+typedef std::pair< uint32_t, size_t > SynapseGID;
+
+/**
+ * Loading of data during SynapsesStream::read(), otherwise load happens
+ * on-demand.
+ */
+enum SynapsePrefetch
```
hernando: Create an enums.h file as in brion?

brain/synapse.cpp
tribal-tec: IIRC this predates circuits from 2012, so I would ignore it for now unless it becomes really necessary to have something here. In the end it would be a duplication of code and computation that has already happened during touch detection.

brain/synapse.h
tribal-tec: It's quite annoying to implement and having an incomplete GID thanks to no extra file for efferent view I'm not convinced yet. I had no use case yes to compare synapses like that.

hernando: Fair enough. Can be left for later.

brain/synapsesIterator.h
```diff
@@ -0,0 +1,56 @@
...
+    BRAIN_API bool operator ==( const SynapsesIterator& rhs ) const;
+    BRAIN_API bool operator !=( const SynapsesIterator& rhs ) const;
+
+    BRAIN_API SynapsesIterator& operator++();
+
+    BRAIN_API Synapse operator*() const;
+
+private:
+    struct Impl;
+    std::unique_ptr< Impl > _impl;
```
hernando: Same comment as for Synapse regarding pimpl.

brain/synapsesIterator.h
```diff
@@ -0,0 +1,51 @@
...
+        public std::iterator< std::forward_iterator_tag, Synapse >
+{
+public:
+    BRAIN_API SynapsesIterator( const Synapses& synapses, size_t index );
+    BRAIN_API ~SynapsesIterator();
+
+    BRAIN_API bool operator ==( const SynapsesIterator& rhs ) const;
+    BRAIN_API bool operator !=( const SynapsesIterator& rhs ) const;
+
+    BRAIN_API SynapsesIterator& operator++();
```
hernando: No postincrement operator?
Not that we really need a more powerful iterator, but have you considered using boost iterator_adaptor? You just need to define a couple of things and inherit from a template and it gives you all the iterator logic plus some typedef to make the iterator usable in some template algorithms and metaprogramming.

brain/synapsesStream.h
```diff
@@ -0,0 +1,85 @@
...
+    BRAIN_API ~SynapsesStream();
+
+    /** @name Move semantics. */
+    //@{
+    BRAIN_API SynapsesStream( SynapsesStream&& rhs );
+    BRAIN_API SynapsesStream& operator=( SynapsesStream&& rhs );
+    //@}
+
+    /**
+     * @return the end of the stream, i.e. any subsequent read() will return
```
hernando: @return if the end of the stream was reached

brain/synapsesStream.h
```diff
@@ -0,0 +1,85 @@
...
+    /** @return the remaining count values for read(). */
+    BRAIN_API size_t getRemaining() const;
+
+    /**
+     * Starts an asynchronous read of the data for the next fraction of synapses
+     * of the requested GIDs.
+     *
+     * @param count the next fraction in the [0,getRemaining()] interval to read
+     * @return a future the Synapses container containing the requested fraction
+     *         of synapses
```
hernando: Review this sentence.

brain/circuit.h
```diff
@@ -120,15 +120,53 @@ class Circuit : public boost::noncopyable
...
+
     /** @return The number of neurons in the circuit. */
     BRAIN_API size_t getNumNeurons() const;
 
+    /**
+     * Access all afferent synapses of the given GIDs.
+     *
+     * @param gids the gids to load afferent synapses for
+     * @param prefetch which synapse data to load on SynapsesStream.read()
+     * @return object to stream synapses data
```
hernando: @return synapse data stream

brain/detail/synapsesStream.h
```diff
@@ -0,0 +1,85 @@
...
+    const Circuit& _circuit;
+    const bool _afferent;
+    const GIDSet _gids;
+    const GIDSet _preGids;
+    const SynapsePrefetch _prefetch;
+    GIDSet::const_iterator _it;
+
+    size_t getRemaining() const
+    {
+        return size_t( std::abs( std::distance( _it, _gids.end( ))));
```
hernando: Why not size_t( std::distance(_gids.end( ), _it ));

tribal-tec: The second iterator must be greater than the first, aka first + distance == second

brain/detail/synapsesStream.h
```diff
@@ -0,0 +1,85 @@
...
+        , _it( _gids.begin( ))
+    {
+    }
+
+    SynapsesStream( const Circuit& circuit, const GIDSet& preGIDs,
+                    const GIDSet& postGIDs, const SynapsePrefetch prefetch )
+        : _circuit( circuit )
+        , _afferent( preGIDs.empty() || (postGIDs.size() < preGIDs.size( )))
+        , _gids( _afferent ? postGIDs : preGIDs )
+        , _preGids( _afferent ? preGIDs : postGIDs  )
```
hernando: If you call the member preGids it's confusing that you may assign the postGIDs.

hernando: Capitalize consistenly (preGIDs vs _preGids).

brain/synapses.cpp
```diff
@@ -0,0 +1,592 @@
...
+    struct FreeDeleter
+    {
+        void operator()( T* ptr ) { free( ptr ); }
+    };
+
+    mutable size_t _size;
+
+    mutable std::unique_ptr< size_t, FreeDeleter<size_t> > index;
+
+    mutable std::unique_ptr< uint32_t, FreeDeleter<uint32_t> > preGID;
```
hernando: typedef std::unique_ptr< uint32_t, FreeDeleter<uint32_t> uint32_ts; ? (and the same for the rest)

brain/detail/circuit.h
```diff
@@ -0,0 +1,535 @@
...
+    const brion::URI _morphologySource;
+    const brion::URI _synapseSource;
+    const brion::URIs _targetSources;
+    mutable brion::Targets _targetParsers;
+    mutable lunchbox::PersistentMapPtr _cache;
+
+    mutable std::unique_ptr< brion::SynapseSummary > synapseSummary;
+    mutable std::unique_ptr< brion::Synapse > synapseAttributes[2];
+    mutable std::unique_ptr< brion::Synapse > synapseExtra;
+    mutable std::unique_ptr< brion::Synapse > synapsePositions[2];
```
hernando: underscores

brain/synapses.cpp
```diff
@@ -0,0 +1,592 @@
...
+                             haveExtra );
+
+        size_t i = 0;
+        for( const auto gid : gids )
+        {
+            const auto& attr = synapseAttributes.read( gid,
+                                                brion::SYNAPSE_ALL_ATTRIBUTES );
+            const auto extra = haveExtra ? synapseExtra->read( gid, 1 )
+                                         : brion::SynapseMatrix();
+            for( size_t j = 0; j < attr.shape()[0]; ++j )
```
hernando: ++j, ++i  remove the increment on i from below

brain/synapses.cpp
```diff
@@ -0,0 +1,592 @@
...
+
+        lunchbox::ScopedWrite mutex( _lock );
+        _loadPositions( _afferent ? _postGIDs : _preGIDs,
+                        _afferent ? _preGIDs : _postGIDs );
+    }
+
+    bool _hasAttributes() const
+    {
+        lunchbox::ScopedRead mutex( _lock );
+        return !!efficacy;
```
hernando: efficacy.get() != 0

brain/synapses.cpp
```diff
@@ -0,0 +1,592 @@
...
+#include <brion/synapse.h>
+#include <brion/synapseSummary.h>
+
+#include <lunchbox/log.h>
+#include <lunchbox/scopedMutex.h>
+
+namespace brain
+{
+
+template<typename T> void allocate( T& data, const size_t size )
```
hernando: Rename to _allocate and enclose in anonymous namespace.

brain/synapses.cpp
```diff
@@ -0,0 +1,592 @@
...
+
+const float* Synapses::preCenterPositionZ() const
+{
+    _impl->_ensurePositions();
+    return _impl->preCenterPositionZ.get();
+}
+
+const uint32_t* Synapses::postGID() const
+{
+    lunchbox::ScopedRead mutex( _impl->_lock );
```
hernando: No mutex needed, this will be available after construction

brain/enums.h
```diff
@@ -0,0 +1,50 @@
...
+
+using namespace brion::enums;
+
+/**
+ * Loading of data during SynapsesStream::read(), otherwise load happens
+ * on-demand.
+ */
+enum SynapsePrefetch
+{
+    SP_NONE = 0,             //!< only loads pre- and post GIDs
```
hernando: I'd prefer keeping the whole type name on the enum prefix.

brain/synapses.cpp
```diff
@@ -0,0 +1,592 @@
...
+        _allocatePositions( haveSize ? _size
+                                     : synapsePositions.getNumSynapses( gids ));
+
+        size_t i = 0;
+        for( const auto gid : gids )
+        {
+            const auto& pos = synapsePositions.read( gid,
+                                                     brion::SYNAPSE_POSITION );
+
+            for( size_t j = 0; j < pos.shape()[0]; ++j )
```
hernando: ++j, ++i

brain/synapses.cpp
tribal-tec: No, if gid is filtered, i shall not be incremented

brain/synapses.cpp
tribal-tec: that's what I did, but shorter :)

brain/synapses.cpp
tribal-tec: Same for preGID, good catch.

brain/synapses.cpp
```diff
@@ -277,8 +281,8 @@ struct Synapses::Impl
             return;
 
         lunchbox::ScopedWrite mutex( _lock );
-        _loadAttributes( _afferent ? _postGIDs : _preGIDs,
-                         _afferent ? _preGIDs : _postGIDs );
+        _loadAttributes( _afferent ? _gids : _filterGIDs,
+                         _afferent ? _filterGIDs : _gids );
```
hernando: This doesn't look right to me. Don't you have to call _loadAttributes( _gids, _filterGIDs ) regardless of the value of _afferent?

hernando: Add a test with lazy loading of the efferent view to check it.

tests/brain/synapses.cpp
```diff
@@ -0,0 +1,221 @@
...
+    const brain::Synapses& synapses =
+            circuit.getAfferentSynapses( circuit.getGIDs( "Layer1" ),
+                                         brain::SYNAPSEPREFETCH_NONE );
+    BOOST_CHECK_EQUAL( synapses.size(), 1172 );
+    BOOST_CHECK_EQUAL( synapses[0].getPresynapticGID(), 10 );
+    BOOST_CHECK_CLOSE( synapses[1].getPostsynapticDistance(), 1.34995711f,
+                       0.00001f );
+    BOOST_CHECK_CLOSE( synapses[2].getConductance(), 0.34758395f, 0.00001f );
+    BOOST_CHECK_EQUAL( synapses[0].getPostsynapticCenterPosition(),
+                brion::Vector3f( 3.799289703f, 1947.041748047f, 9.237932205f ));
```
hernando: Compare the data to another container loaded with prefetching, that makes the test more resilient to changes.

hernando:
+1

PR #88, opened 2016-08-25
Enums classes and functions renamings

hernando:


brain/neuron/morphology.cpp
```diff
@@ -89,7 +89,7 @@ const Vector2is& Morphology::getSections() const
     return *_impl->sections;
 }
 
-const SectionTypes& Morphology::getSectionTypes() const
+const brion::SectionTypes& Morphology::getSectionTypes() const
```
tribal-tec: I would keep the using in types.h to avoid the brion:: here.

brain/neuron/morphology.cpp
```diff
@@ -125,7 +125,7 @@ Sections Morphology::getSections( const SectionTypes& types ) const
 
 Section Morphology::getSection( const uint32_t& id ) const
 {
-    if(( *_impl->types )[id] == SECTION_SOMA )
+    if(( *_impl->types )[id] == brion::enums::SECTION_SOMA )
```
tribal-tec: dito

brain/neuron/morphologyImpl.cpp
```diff
@@ -36,7 +36,7 @@ namespace neuron
 namespace
 {
 template< typename T >
-void _serializeArray( uint8_t*& dst,
+void _serializeArray( unsigned char*& dst,
```
tribal-tec: why changed?

brain/neuron/morphology.cpp
tribal-tec: Ok, why not using the 'new' values?

brain/neuron/types.h
```diff
@@ -32,10 +33,20 @@ class Morphology;
 class Section;
 class Soma;
 
+enum class SectionType
+{
+    soma = brion::enums::SECTION_SOMA,
+    axon = brion::enums::SECTION_AXON,
+    dendrite = brion::enums::SECTION_DENDRITE,
+    apicalDendrite = brion::enums::SECTION_APICAL_DENDRITE,
```
tribal-tec: apical_dendrite to help python?

brain/neuron/morphology.cpp
hernando: I don't like that very much, because then you have brain::SectionTypes and brain::morphology::neuron::SectionTypes.

doc/Changelog.md
```diff
@@ -3,6 +3,8 @@ Changelog {#Changelog}
 
 # Release 1.9.0 (git master)
 
+* [88](https://github.com/BlueBrain/Brion/pull/88):
+  Brain namespace enums made strongly typed.
```
tribal-tec: plus fixing the bug in the morphology class

brain/neuron/morphologyImpl.cpp
hernando: I read in SO that uin8_t is less safe than unsigned char. In some architectures uin8_t could be an extended numerical type for efficiency (the standard allows it) while unsigned char is always guaranteed to be the minimum addressable information unit (I wonder if it can be different from a byte nowadays). Also strict aliasing rules are different for unsigned char*, unsigned char* can alias any other type for uin8_t* the normal strict alias rules apply.

On the other hand I've just remembered and double checked that we are using -fno-strict-aliasing. We should remove that option and fix the code that doesn't work because the performance impact can be non-negligible, and we are not using __restrict__ pervasively to mitigate it.

brain/neuron/types.h
hernando: For the same reason we should rename all the functions then :)

brain/neuron/morphology.cpp
hernando: Because then I have to cast ( *_impl->types )[id] to the new enum. I don't want to use the new type in the member variable because that requires an interpret cast of shared_ptr<vector<SectionTypes> > from brion enum to brain enum or copying the whole array.
It really annoys me how it looks now, but that's the best that I could find.

brain/enums.h
```diff
@@ -26,26 +26,20 @@
...
-{
-
-using namespace brion::enums;
 
 /**
  * Loading of data during SynapsesStream::read(), otherwise load happens
  * on-demand.
  */
-enum SynapsePrefetch
+enum class SynapsePrefetch
```
tribal-tec: move to types.h

PR #92, opened 2016-09-07
Varargs be damned (Fix clang/OS X build)

eile:


tests/brain/morphology.cpp
```diff
@@ -347,16 +319,15 @@ BOOST_AUTO_TEST_CASE( transform_with_matrix )
     brain::Matrix4f matrix;
     matrix.rotate_z( M_PI * 0.5 );
     brain::neuron::Morphology rotated( TEST_MORPHOLOGY_URI, matrix );
-    checkCloseArraysUptoN( rotated.getPoints(), 4,
-      V4f( .0, .1, .0, .1 ), V4f( -.1, .0, .0, .1 ),
-      V4f( .0, -.1, .0, .1 ), V4f( .1, .0, .0, .1 ));
+    checkCloseArraysUpToGiven( rotated.getPoints(),
+                               { V4f( .0, .1, .0, .1 ), V4f( -.1, .0, .0, .1 ),                                  V4f( .0, -.1, .0, .1 ),V4f( .1, .0, .0, .1 )});
```
hernando: missing line jump

hernando:
You could have created a branch in Brion. That would have made it easier for me to reproduce the build error in Ubuntu :)

hernando:
I couldn't reproduce the build error on my 16.04, any idea?

eile:
Parsing https://gcc.gnu.org/bugzilla/show_bug.cgi?id=28901

hernando:
I remember seeing that error before, but only when there were other compilation errors, which is not the case. Whatever is the reason, what doesn't make sense at all to me is that I don't get the warning using the same compiler and same build type.
If adding LB_UNUSED is the best solution let's go with it. Not that I like it, but at least I don't have to look at that code very often.

Issue #90, opened 2016-08-29
morphologyConverter creates invalid h5 file from SWC


juanchopanza:
The morphologyConverter tool creates h5 files with more than one soma section when given an SWC file with neurites branching out from soma points other than the final soma point. These are commonly used layouts:

* For example, the so-called "contour" soma. A is represented as a line of soma points, with neurite branches coming out of arbitrary points. Topologically, this forms multiple soma sections, but in H5 there is supposed to be only one soma section. 

* Neurites branching out of the initial soma point.

It isn't entirely clear how to deal with these cases, because the single soma section restriction implies some loss of information (but one could, for instance, add the branching points as initial points to the neurite sections.) In any case, it is not good to create invalid h5 files from valid SWC ones.

To reproduce, on bbpviz1.cscs.ch:

```
module load BBP/viz/2016.R2
morphologyConverter -i nrn.swc -f h5v1
```
where `nrn.swc` is this:

```
# Soma with neurites branching out of the middle
# This is the layout of the "contour soma".
1 1 0 0 0 0.0 -1
2 1 0 0 2 0.0 1
3 1 0 0 3 0.0 2
4 1 0 0 4 0.0 3
5 1 0 0 5 0.0 4
6 1 0 0 6 0.0 5
7 1 0 0 7 0.0 6
8 1 0 0 8 0.0 7
# first neurite branch, from point 3
9 2 1 0 3 0.5 3
10 2 2 0 3 0.5 9
11 2 3 0 3 0.5 10
12 2 4 0 3 0.5 11
13 2 5 0 3 0.5 12
# second neurite branch, from point  6
14 3 0 1 6 0.5 6
15 3 0 2 6 0.5 14
16 3 0 3 6 0.5 15
17 3 0 4 6 0.5 16
18 3 0 5 6 0.5 17
# third neurite branch, from point 8
19 4 0 0 9 0.5 8
20 4 0 0 10 0.5 19
21 4 0 0 11 0.5 20
22 4 0 0 12 0.5 21
23 4 0 0 13 0.5 22
```

The same can be observed with a morphology with neurites branching out of the initial soma point, for example

```
# Soma with neurites branching out of the initial point
1 1 0 0 0 0.0 -1
2 1 0 0 2 0.0 1
3 1 0 0 3 0.0 2
4 1 0 0 4 0.0 3
5 1 0 0 5 0.0 4
6 1 0 0 6 0.0 5
7 1 0 0 7 0.0 6
[...]

PR #94, opened 2016-09-19
Fix for #90: SWC parser creates multiple soma sections.

hernando:
This happened when neurites were connected to arbitrary soma samples.

### DIFF
--- a/brion/plugin/morphologySWC.cpp
+++ b/brion/plugin/morphologySWC.cpp
@@ -348,8 +348,14 @@ void MorphologySWC::_buildSampleTree( RawSWCInfo& info )
 
         if( !root )
         {
-            Sample& parent = samples[sample.parent];
-            if( !parent.valid )
+            if( sample.parent == int( currentSample ))
+            {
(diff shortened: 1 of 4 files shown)


hernando:
Just a minor comment. Notice that the current parser does two different things when the soma is given as a contour and as a point. When the soma is a point, the first order sections are not connected to that point (that would be wrong for several use cases: distance to soma computation, simple sphere + tubes rendering, ...). On the other hand, when the soma is described as a contour (or a tree), then the neurites are connected to the soma samples. This may not be always correct, but when that's the case the swc file should be fixed because that's how I interpret that case.

hernando:
The SWC ans H5 are already not fully equivalent and it's not possible to to SWC->H5->SWC and get the original result (that's not a big deal since there's no SWC writer anyway).
With this commit, the behaviour for single point somas has not changed compared to before. In that case the first point of a first order section is the point which has the soma sample as parent, i.e. the section is not extended adding the soma point as the first one.

PR #100, opened 2016-10-03
Resolve MVDTool via find_package() for Nix support

tribal-tec:


tribal-tec:
retest this please

hernando:
No that I realize, shouldn't you add MVDTool to target_link_libraries even if it's header only.

tribal-tec:
It's there in brain/CMakeLists.txt:51

hernando:
Yes you're right.

tribal-tec:
retest this please

PR #103, opened 2016-10-13
Change synapse prefetching default to none.

hernando:


tests/brain/python/circuit.py
```diff
@@ -62,7 +62,7 @@ def test_load_morphology(self):
 
     def test_metypes(self):
         count = self.circuit.num_neurons()
-        mnames = self.circuit.morphology_names()
+        names = self.circuit.morphology_names()
```
tribal-tec: sure?

tribal-tec:
remove the brain::SynapsePrefetch::none from the unit tests

hernando:
I've replaced none with all which is the case that was not covered after the default parameter was changed.

tribal-tec:
Hmm, I didn't mean that. In the C++ tests which have explicit 'none', those ones can be changed to remove the now obsolete 'none' and use 'all' for those which rely on the not-true anymore 'all'.

hernando:
I thought you were referring to the Python tests.

hernando:
ping

PR #104, opened 2016-10-13
Changes for Eyescale/Lunchbox#278: Simplified Plugin API

eile:


CMakeLists.txt
```diff
@@ -19,14 +19,14 @@ set(BRION_MAINTAINER
...
+  libboost-test-dev libboost-thread-dev)
 
 set(COMMON_PROJECT_DOMAIN ch.epfl.bluebrain)
 include(Common)
 
 common_find_package(BBPTestData)
-common_find_package(Boost REQUIRED COMPONENTS date_time filesystem
-                    program_options regex system thread unit_test_framework)
+common_find_package(Boost REQUIRED COMPONENTS
+  date_time filesystem program_options regex system thread unit_test_framework)
```
hernando: I was about to comment that we hadn't made this commit on top of master when you just changed it.

PR #106, opened 2016-10-24
Add data model documentation for compartments and spikes

tribal-tec:


doc/datamodel/compartments.md
```diff
@@ -0,0 +1,53 @@
...
+The branches of the tree are formed by _sections_ which are subdivided into
+smaller pieces. They start at the soma and eventually branch into multiple
+sections in a tree-like structure.
+
+During the reconstruction from real neurons, each section is formed by a number
+of _segments_. Each segment has two endpoints in 3D space, as well as a radius
+for each endpoint. These segments consequently represent tubelets in 3D space.
+
+For the simulation, each section is decomposed into m _compartments_, which have
+a n : m mapping to segments. For each compartment, various observables are
```
hernando: I don't think there's any well defined mapping from compartments to segments.

doc/datamodel/compartments.md
```diff
@@ -0,0 +1,53 @@
...
+simulated and may be reported in a CompartmentReport.
+
+## Meta data
+
+* list of neuron GIDs
+* start time in ms
+* end time in ms
+* dt in ms
+* data unit as string, e.g. mV, pA
```
hernando: Where is this metadata coming from, the report itself or the blue config?
The blue config also has the forward skip parameter.

doc/datamodel/compartments.md
```diff
@@ -0,0 +1,53 @@
...
+## Compartment information
+
+* for each neuron, sorted ascending by GID
+* number of compartments of each section per neuron, sections order defined by
+  morphology
+* all sections of a neuron are always described, i.e. unreported sections have
+  zero compartments
+
+Compartments are always linearly interpolated along all segments of a section,
+i.e. compartments have all the same length within a particular section.
```
hernando: Remove the sentence about linear interpolation and keep the sentence after i.e. I don't what linear interpolation has to do with splitting a section in equally sized parts.

doc/datamodel/compartments.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Report information
+
+* Reportable value in data unit per compartment
+* Sorted ascending by timestamp, discretized in (_endtime - starttime_) / _dt_
+  frames
+* Within a timestamp, sorted ascending by GID
```
hernando: I'm not sure this is true.

doc/datamodel/spikes.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Meta data
+
+* list of neuron GIDs
+* start time in ms
+* end time in ms
+
+## Report information
+
+* timestamp in ms to GID tuples or vice versa, unordered
```
hernando: Pairs of timestamp in ms and GID or vice versa

doc/datamodel/compartments.md
```diff
@@ -0,0 +1,53 @@
...
+During the reconstruction from real neurons, each section is formed by a number
+of _segments_. Each segment has two endpoints in 3D space, as well as a radius
+for each endpoint. These segments consequently represent tubelets in 3D space.
+
+For the simulation, each section is decomposed into m _compartments_, which have
+a n : m mapping to segments. For each compartment, various observables are
+simulated and may be reported in a CompartmentReport.
+
+## Meta data
+
```
hernando: You always start enumerations with lowercase and omit dots at the end of each line. Is that proper English punctuation? Because it looks strange to me, at least when each item is a sentence.

eile:
> what is the scope and expected level of detail of this document?

To describe the data model of our simulation data. It is precisely not a spec for the file format.

eile:
> also giving some details about the file layout

Where?

doc/datamodel/compartments.md
eile: According to @jamesgking, we can assume a linear interpolation between the n segments to the m compartments of one section.

hernando: When I read this sentence it's not very clear what "n : m mapping to segments means" the letters n and m even seem reversed. 
I would keep the language simple and clear and use mathematical language only to avoid the ambiguity of natural language, but then the notation and concepts need to be clear. I'd simply say: For the simulation, each section is subdivided in m equally long _compartments_. Which segments are overlapped by a compartment is simply a corollary of this definition which doesn't need any explanation, but if you want to define it mathematically it gets complicated (it's not as simply as saying n : m mapping).
Regarding the linear interpolation I have a comment below.

doc/datamodel/compartments.md
hernando: I really don't understand it, to me the sentences are not equivalent.

What does linear interpolation mean for you here? To me interpolating is finding a function which is constrained to pass through a set of given sample points (which typically come from a continuous function we want to approximate but is not fully known). What's the function being interpolated here, what are the control points?

Could Jim mean that since the variables measured at compartments are continuous, compartments could be understood as sample points that are equally spaced along a section (i.e., the sample points are placed at the center of the compartments), and that you could do linear inteporlation along a section to find the approximate value of the variable at the segments?

hernando:
For example: "timestamp in ms to GID tuples or vice versa, unordered", what you really have is a set, the fact that the set is enumerated in any order in the file is an implementation detail, the mathematical concept stays the same. Sorted by ascending GIDs is another example.

PR #113, opened 2016-11-28
No surface positions

hernando:


brain/detail/circuit.h
```diff
@@ -354,19 +362,22 @@ class Circuit::Impl
...
                 // size in bytes (although there's a getter for it used in
                 // saveSynapsePositionsToCache()), so we reconstruct the row and
                 // column count here.
-                const size_t numColumns = brion::SYNAPSE_POSITION_ALL;
-                const size_t numRows = size / sizeof(float) / numColumns;
-                brion::SynapseMatrix values( boost::extents[numRows]
-                                                           [numColumns]);
+                const size_t numColumns = _synapsePositionColumns;
+                const size_t numRows =
+                    size / sizeof(float) / numColumns;
```
rdumusc: style: keep on one line

brain/detail/circuit.h
```diff
@@ -270,6 +271,13 @@ class Circuit::Impl
             _synapsePositions[i]->reset(
                 new brion::Synapse( _synapseSource.getPath() + (afferent ?
                       afferentPositionsFilename : efferentPositionsFilename )));
+
+        if (_synapsePositionColumns == 0)
```
rdumusc: style: spaces inside (), use {} for if() statement longer than one line, or maybe define
const auto& synapse = *_synapsePositions[i];
before these lines

brion/synapse.cpp
```diff
@@ -357,20 +395,17 @@ class Synapse : public boost::noncopyable
         lunchbox::ScopedWrite mutex( detail::_hdf5Lock );
         SilenceHDF5 silence;
 
-        BOOST_FOREACH( UnmappedFiles::value_type& entry, _unmappedFiles )
+        // this trial-and-error is the 'fastest' path found
+        BOOST_FOREACH( const std::string& candidate, _fileNames )
         {
-            const std::string& candidate = entry.first;
-            H5::H5File& file = entry.second;
-            // keeping the files open 'saves' some time
```
rdumusc: If this comment still holds true, keep it

hernando: No, it's not. I changed the code to not keep the files open because otherwise the process will run out of file descriptors.

brain/detail/circuit.h
hernando: done

brain/detail/circuit.h
hernando: done

brion/synapse.cpp
rdumusc: Yes, indeed it seemed like it did ;-)
It might be too much work to do this now, but I think this modification would have deserved to be in a separate commit with this explanation in the commit message. It helps for future reference and for the code review to know why things have been modified.

rdumusc:
Otherwise +1 for me

brion/synapse.cpp
hernando: You're right. I will modify the commit message at least.

hernando:
I need a positive review to merge

PR #119, opened 2016-12-15
Optimize brain::Circuit::loadMorphologies

eile:


brain/circuit.cpp
```diff
@@ -98,19 +101,15 @@ neuron::Morphologies Circuit::loadMorphologies( const GIDSet& gids,
     GIDSet::const_iterator gid = gids.begin();
     for( size_t i = 0; i < uris.size(); ++i, ++gid )
```
favreau: Can you use C++11 syntax there? for( const auto& gid: gids)

hernando: No because there are counters being incremented in the loop

favreau: I know but simply move ++gid inside the loop, and that gets rid of the GIDSet::const_iterator. Anyway, just a suggestion there.

hernando: Up to @eile to decide, I tend to avoid increments inside the loop body. On the other hand: auto gid = gids.begin(); is welcome.

PR #123, opened 2017-01-06
Fix naming of map-based compartment reports

eile:


apps/compartmentConverter.cpp
```diff
@@ -162,9 +163,22 @@ int main( const int argc, char** argv )
...
+    {
+        try
+        {
+            outURI.setPath( boost::filesystem::canonical(
+                                inURI.getPath( )).generic_string( ));
+        }
+        catch( const boost::filesystem::filesystem_error& )
+        {
+            // For non-filebased reports, the canonical above will throw.
+            outURI.setPath( inURI.getPath( ));
```
rdumusc: I don't think I understand what this path setting is about. Is it trying to place the output file in the same folder as the input file? But then if the inURI is not a file, and the above throws, what is the point of the second line? Won't it be empty anyway?

eile: If your output is a file, you will naturally give it a path so you won't get here.

If your output is a key-value store, reuse the file-path as the "path" for the KV-store. The KV-store uses the path to uniquefy the keys so that reports don't overwrite each other.

If your input is a file-based report, canonical will make sure that the same file gets written to the same output path. If it's not file-based, the "path" will throw and is then assumed to be "absolute".

rdumusc: Thinking a bit more it must be something different. Like reading from some uri with a schema: memcached://myserver1/somefile and outputing to memcached://myserver2/ without an explicit path. I am getting closer? The description of the main() still says: "Convert a compartment report to an HDF5 report" and could no doubt be updated :-)

eile: BTW, this is the reason for the Plugin::getDescription change I am working on ;)

eile: Yes, you hit the nail.

rdumusc: Yes thanks, I got confused between the path in the uri sense and the path in the filesystem sense (folder), especially because I saw that the default "out.h5" didn't have a "path"... that's still quite convoluted ;-)
And BTW, I know it's not directly what you were working on, but I think ( "output,o", po::value< std::string >(), "Output report URI" ) could be: ( "output,o", po::value< std::string >()->default_value("out.h5"), "Output report URI" )

eile: don

PR #124, opened 2017-01-09
Descriptions for plugins

eile:

PR #125, opened 2017-01-11
Fix C&P error in #124

eile:


brion/spikeReport.cpp
```diff
@@ -36,36 +38,37 @@ const std::string spikePluginDSONamePattern( "Brion.*SpikeReport" );
...
-                                                    lunchbox::getLibraryPaths(),
-                                                    spikePluginDSONamePattern );
-        }
-    };
-
     explicit SpikeReport( const SpikeReportInitData& initData )
     {
-        static DSOPluginsLoader loader;
-        plugin.reset( SpikePluginFactory::getInstance().create( initData ));
+        _loadPlugins();
```
hernando: I think the previous code was thread-safe with regard to multiple threads creating the first SpikeReport at the same time because static variable initialization is thread safe (but I'm not 100% sure). The new version is not, or it it?

eile: o_O

eile: Fixed

eile:
> Smells like clang-format.

Indeed.

> I think you should've restrained from applying it before everybody accepts the style.

How would I (and you) gather evidence on what is good or not with clang-format?  I'ld rather start using it and getting feedback to see the style changes.

hernando:
Well, I see something. There's too much whitespace around parentheses for my taste, but if everybody is half happy with this I can live with it, the rest of the style changes look OK to me.

As I understand it then, you'd rather go for a period of experimentation and fix the style of the whole project once and for all later?

eile:
Better? :P

eile:
> whitespace around parentheses

With clang-format it's either that, or whitespace outside: ```if (foo && bar)```, unless I missed something

eile:
> go for a period of experimentation and fix the style of the whole project once and for all later

The only constant in the universe is change. ;)

I did not have a plan like that, I simply felt without trying it I won't get enough data to decide.

eile:
F! clang-format was unintentionally still on, fixing.

PR #128, opened 2017-01-24
Partial SpikeReport code review and improvements

eile:


brion/plugin/spikeReportBinary.cpp
```diff
@@ -159,25 +132,22 @@ bool SpikeReportBinary::handles( const SpikeReportInitData& initData )
...
 
 Spikes SpikeReportBinary::read( const float )
 {
     // In file based reports, this function reads all remaining data.
     Spikes spikes;
-    Spike* spikeArray = _memFile.getSpikes();
-    const size_t nElems = _memFile.getSpikeCount();
+    const Spike* spikeArray = _memFile->getReadableSpikes();
+    const size_t nElems = _memFile->getNSpikes();
+    _currentTime = spikeArray[nElems - 1].first +
```
kaabimg: In the case of an empty spike report file (only a header, with no spikes, this will access to an empty array

brion/plugin/spikeReportBinary.cpp
```diff
@@ -199,24 +169,22 @@ Spikes SpikeReportBinary::readUntil( const float max )
         pushBack( spikeArray[_startIndex], spikes );
     }
 
-    if ( _startIndex == nElems )
+    if( _startIndex == nElems )
     {
-        _currentTime = UNDEFINED_TIMESTAMP;
+        _currentTime = spikeArray[nElems - 1].first +
```
kaabimg: Same comment as above : access to an empty array when _startIndex == nElems == 0

brion/plugin/spikeReportBinary.cpp
```diff
@@ -38,113 +38,86 @@ namespace
...
+    {
+        return reinterpret_cast< Spike* >( _map.getAddress< uint8_t >() +
+                                           sizeof( Header ));
+    }
 
-bool BinaryReportFileMemMap::isMappedForRead() const
-{
-    return _mappedForRead;
-}
+    operator bool() const { return _map.getAddress() != nullptr; }
```
kaabimg: This operator is no longer used

eile: removed

brion/plugin/spikeReportBinary.cpp
eile: done

brion/plugin/spikeReportBinary.cpp
eile: done

eile:
Nice tumbleweed here.

brion/spikeReport.h
```diff
@@ -62,7 +62,7 @@ class SpikeReport : public boost::noncopyable
      * State::ended means that the report end is reached (in read mode)
      * State::failed means that a error occured when readyn or writing the report
      */
-    enum class State : uint8_t
+    enum class State : uint32_t
```
hernando: What's the reason for using uint32_t? (in this case I think you can even omit it)

brion/plugin/spikeReportBinary.cpp
```diff
@@ -236,26 +210,22 @@ void SpikeReportBinary::readSeek( const float toTimeStamp )
...
 
     if ( spikes.empty() )
         return;
 
     // create or resize the file
-    if ( !_memFile || _memFile.getSpikeCount() != totalSpikeCount )
-        _memFile.mapForReadWrite( getURI().getPath(), totalSpikeCount );
-
-    Spike* spikeArray = _memFile.getSpikes();
+    if ( _memFile->getNSpikes() != totalSpikes )
```
hernando: I preferred getSpikeCount, I don't like this sort of abbreviations.

brion/plugin/spikeReportBinary.cpp
```diff
@@ -55,85 +53,68 @@ class Header
...
-Spike* BinaryReportFile::getSpikes() const
-{
-    return reinterpret_cast< Spike* >( _map.getAddress< uint8_t >() +
-                                       sizeof( Header ));
-}
+    Spike* writeSpikes()
+    {
+        return reinterpret_cast< Spike* >( _map.getAddress< uint8_t >() +
+                                           sizeof( Header ));
+    }
```
hernando: Rename to mutableSpikes() and constSpikes() or use getSpikes for both with const and mutable overloads.

brion/plugin/spikeReportBinary.cpp
```diff
@@ -90,13 +90,13 @@ class BinaryReportMap
...
 
-    const Spike* readSpikes() const
+    const Spike* getReadableSpikes() const
     {
         return reinterpret_cast< const Spike* >( _map.getAddress< uint8_t >() +
                                                  sizeof( Header ));
     }
 
-    Spike* writeSpikes()
+    Spike* getWritableSpikes()
```
hernando: OK, I was looking at the commits in order to make it easier to digest.

brion/plugin/spikeReportBinary.cpp
```diff
@@ -137,13 +140,14 @@ Spikes SpikeReportBinary::read( const float )
...
+    const Spike* spikeArray = _memFile->getReadableSpikes();
     const size_t nElems = _memFile->getNSpikes();
+    _currentTime = spikeArray[nElems - 1].first +
+                   std::numeric_limits< float >::epsilon();
 
     for ( ; _startIndex < nElems; ++_startIndex )
         pushBack( spikeArray[_startIndex], spikes );
 
-    _currentTime = UNDEFINED_TIMESTAMP;
     _state = State::ended;
```
hernando: State shouldn't be ended neither.

brion/plugin/spikeReportASCII.cpp
```diff
@@ -48,7 +46,8 @@ Spikes SpikeReportASCII::read( const float )
     Spikes spikes;
     auto start = _lastReadPosition;
     _lastReadPosition = _spikes.end();
-    _currentTime = UNDEFINED_TIMESTAMP;
+    _currentTime = _spikes.rbegin()->first +
+                   std::numeric_limits< float >::epsilon();
```
hernando: The spec doesn't say anything about the value of getCurrentTime() when state is ended, it's undefined. For that reason, your code shouldn't rely on any specific value. My gut tells me this change is wrong if you don't constraint the specification, but referring to deltas there also seems wrong. Do you really need this change to make the incremental converter work?

hernando: The same goes for all other places in which you've done the same.

brion/plugin/spikeReportBinary.cpp
```diff
@@ -115,6 +115,9 @@ SpikeReportBinary::SpikeReportBinary( const SpikeReportInitData& initData )
         _memFile.reset( new BinaryReportMap( getURI().getPath( )));
     else
         _memFile.reset( new BinaryReportMap( getURI().getPath(), 0 ));
+
+    if( _memFile->getNSpikes() > 0 )
+        _currentTime = _memFile->getReadableSpikes()[0].first;
```
hernando: This is the only place I find really acceptable to replace -inf by something else, but you have to update the specification of getCurrentTime accordingly

brion/plugin/spikeReportASCII.cpp
```diff
@@ -46,8 +46,7 @@ Spikes SpikeReportASCII::read( const float )
     Spikes spikes;
     auto start = _lastReadPosition;
     _lastReadPosition = _spikes.end();
-    _currentTime = _spikes.rbegin()->first +
-                   std::numeric_limits< float >::epsilon();
+    _currentTime = UNDEFINED_TIMESTAMP;
```
hernando: Argh, this incremental reviewing approach doesn't seem to be a good idea...

I think you should make separate PRs for independent things, for example the command line argument parsing could have gone in a separate one.

brion/spikeReport.h
eile: uint8_t seemed like a premature optimization. Will remove type.

apps/CMakeLists.txt
```diff
@@ -1,26 +1,27 @@
...
 set(COMPARTMENTCONVERTER_SOURCES compartmentConverter.cpp)
 set(COMPARTMENTCONVERTER_LINK_LIBRARIES Brion ${Boost_PROGRAM_OPTIONS_LIBRARY})
+
 if(TARGET BBPTestData)
   list(APPEND COMPARTMENTCONVERTER_LINK_LIBRARIES BBPTestData)
+  list(APPEND SPIKECONVERTER_LINK_LIBRARIES BBPTestData)
 endif()
 
+common_application(spikeConverter)
+common_application(morphologyConverter)
```
eile: It allows grouping the link libs in if(testdata) above.

brion/plugin/spikeReportBinary.cpp
eile: Renamed to getNumSpikes to be consistent with the rest of the API

brion/plugin/spikeReportBinary.cpp
eile: are you fine with the current getWritableSpikes/getReadableSpikes?

hernando: It's fine. Using the same name for both doesn't work?

PR #129, opened 2017-01-27
Simplify ASCII report implementation

eile:
@hernando, @rdumusc, @kaabimg: Whenever you are done with an
implementation, please review it again with an eye to make it
simpler. This change is a good example of the cruft which will
accumulate otherwise to a state which nobody will eventually
understand. My initial trigger was 'spikeReportFile and ASCII? WTH?':

* srFile was a class which was created, one function called, done ->
  make this a free function
* srFile::write was dead code
* the code had an intermediate map representation for no reason at all

The last change brings the read time to 2s instead of 2.7s for a 33MB
report I had at hand.

@kaabimg: Do not forget the ABI version whenever you change the API or
ABI of a plugin, otherwise one gets random crashes when incompatible
plugins are loaded.

Happy weekend!


hernando:
@eile I see your point, but this is not writing an implementation, it was modifying an already existing one. In this case it's difficult to say to which extent the original PR should have dropped existing code. It already took long enough to get reviewed without addressing all the possible flaws in  code that was otherwise working as expected. In other occasions I've received as answer to a requested change of this type in a PR that it was feature creep.
For the sake of readability of pull requests I'd defend what was done in this particular case, as it's clear that the changes that you're commenting are perfectly fine as a follow up PR.

apps/spikeConverter.cpp
```diff
@@ -62,16 +62,16 @@ int main( int argc, char* argv[] )
...
     positional.add("input", 1);
-    positional.add("output", 2);    
-    
+    positional.add("output", 2);
+
     po::variables_map vm;
     try
     {
         po::store( po::command_line_parser( argc, argv ).
-                   options( options ).positional( positional ).run(), vm );
```
hernando: Hmm, the web editor doesn't highlight trailing whitespace.

apps/spikeConverter.cpp
```diff
@@ -93,6 +93,13 @@ int main( int argc, char* argv[] )
                   << brion::Version::getString() << std::endl;
         return EXIT_SUCCESS;
     }
+    if( vm["input"].as< std::string >() == vm["output"].as< std::string >( ))
+    {
+        std::cerr << "Cowardly refusing to convert "
+                  << vm["input"].as< std::string >() << " onto itself"
+                  << std::endl;
+        return EXIT_FAILURE;
+    }
```
hernando: Since you're touching this code, I'd remove all default values for input and output. These values are completely useless in my opinon, furthermore, after I rescued the positional arguments it took me some time to realize that the converter was not working as before and that positional arguments were ignored (good that I used the tool from your PR for a test and I realized this change in behaviour, because it'd have been bad to merge it). If the tool is run without sufficient parameters, it should print an error and help, not do some default non-informative useless action.

brion/plugin/spikeReportASCII.h
```diff
@@ -45,6 +44,12 @@ class SpikeReportASCII : public SpikeReportPlugin
 protected:
     Spikes _spikes;
     Spikes::iterator _lastReadPosition;
+
+    // Returns true if parsing succeeded
+    using ParseFunc = std::function< bool( const std::string&, Spike& ) >;
+
+    static Spikes parse( const Strings& files, const ParseFunc& parse );
+    static Spikes parse( const std::string& file, const ParseFunc& parse );
```
hernando: The same approach  should be followed for file writing. In the current implementation SpikeReportNEST and SpikeReportBluron duplicate all the code but one line.

apps/spikeConverter.cpp
eile: I disagree: by default, it converts a testdata ascii report to a binary format. In particular, using  out.spikes as default for output is what one wants in 80% of the cases, so make simple things simple.

Not sure what you mean with the positional reasoning. Imo that you could do 'spikeConverter in out' was just laziness on the implementers part (yours truly), and the -i/-o is now consistent with the compartmentConverter.

apps/spikeConverter.cpp
eile: clang-format.

brion/plugin/spikeReportASCII.h
eile: Ack

eile: I think that used to be the case before the last iteration on the spike report API, so it was missed during review there. ;)

brion/plugin/spikeReportASCII.cpp
```diff
@@ -192,7 +192,7 @@ void _parse( Spikes& spikes, const std::string& filename,
                                         filename + " failed at line " +
                                         std::to_string( lineNumber )));
        }
-       spikes.push_back( spike );
+       spikes.emplace_back( spike );
```
hernando: Does it make a difference with a copyable object that you're not moving (although in this case even moving and copying are the same).

apps/spikeConverter.cpp
hernando: You're right about the output parameter, I didn't consider that, but I disagree for the input, I couldn't care less about the test data when using this tool.
Regarding being able to simply do "spikeConverter in out" it's not laziness at all, positional arguments for necessary arguments leaving an optional one with a default at the end it's a pretty common CLI pattern. Many tools work like that, to name a few: cp, mkdir, cat, convert, grep, even for gcc you can do "gcc file.cpp".

But the most important detail is that removing positional arguments was a breaking change.

apps/spikeConverter.cpp
hernando: I did the edition inside Github itself. Could I have pushed on top of your branch from the command line?

brion/plugin/spikeReportASCII.cpp
```diff
@@ -213,14 +213,36 @@ Spikes SpikeReportASCII::parse( const Strings& files, const ParseFunc& parse )
...
+    if ( !file.is_open() )
+    {
+        _state = State::failed;
+        return;
+    }
+
+    for ( const Spike& spike : spikes )
+        writefunc( file, spike );
+
+    file.flush();
```
eile: Why is this flush needed? file dtor will be called soon and should flush?

PR #131, opened 2017-02-02
Added SpikeReport::getEndTime and supportsBackwardsSeek

hernando:
Enable caching of spikes in SpikeReportReader for reports that don't support
backward seek.
Fix SpikeReportReader wrapping with latest API.
- Make getSpikes return a numpy array with dtype "f4, u4" for spikes.
- Added docstrings.

### DIFF
--- a/brion/spikeReport.h
+++ b/brion/spikeReport.h
@@ -161,6 +161,16 @@ public:
     BRION_API float getCurrentTime() const;
 
     /**
+     * @return the end timestamp of the report. This is the timestamp of the
+     *         last spike known to be available or written or larger if the
+     *         implementation has more metadata available.
+     *         For stream reports this time is 0 before any operation is
+     *         completed.
+     * @version 2.0
+     */
+    BRION_API float getEndTime() const;
+
+    /**
      * @return The state after the last completed operation.
      * @version 2.0
      */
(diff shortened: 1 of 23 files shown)


hernando:
The test is failing due to precision error in double to float conversion of viceversa. It'd be good if you can take a look before I'm back.

eile:
Update spec please

brion/spikeReport.h
```diff
@@ -279,6 +287,12 @@ class SpikeReport : public boost::noncopyable
      */
     BRION_API void write( const Spikes& spikes );
 
+    /**
+     * @return Whether the report supports seek to t < getCurrentTime() or not.
+     * @version 2.0
+     */
+    BRION_API bool supportsBackwardSeek() const;
+
```
eile: This looks maeh. Why not have a bool return value for seek()?

kaabimg: Seek returns a future and throws if backward seek is not supported. It would be better to know in advance if the feature is available or not.

hernando:
Fivox tests pass again with these changes.

Do you want me to update spikeReport.md? I thought this spec files were RFCs that didn't need to be kept up to date with the real documentation.
The changes I made in SpikeRepor are small, I think the code and the commit messages are sufficient to understand why they were made.

brion/spikeReport.h
hernando: This function is really needed. Before you even seek you have to decide whether you're going to store the collected spikes or not because moving forward works in all cases but you still want to keep the spikes for future random access. Waiting until seek returns false (or throws) to start storing means that you may have already lost all the data (e.g. if you let playback move forward until the end and you want to loop it over afterwards).

PR #132, opened 2017-02-07
Improve write() test coverage and fix behaviour

eile:


brion/spikeReport.h
```diff
@@ -53,7 +53,7 @@ class SpikeReport;
  *
  * This class is not thread-safe except where stated otherwise.
  */
-class SpikeReport : public boost::noncopyable
+class SpikeReport
```
kaabimg: I would prefer to keep the boost::noncopyable: It makes clear from the beginning that the class is not copyable and I don't have to scroll down to the private section to understand that.
The noncopyable boost header is anyway included from <types.h>

tribal-tec: Imo we removed it in other places in favor of the C++11 support for =delete. For readability one could put the private =delete section at the top of the class definition.

kaabimg: I get this point, but sometimes =delete is used for some but not all constructors/assignment operators in order. In this case you will have to check all of them to check if the type is noncopyable or not. For me, boost::noncopyable is much clearer and makes the code shorter.

eile: Right, but I just made this class movable.

hernando: Certainly there's nothing wrong to make this class movable as iostreams are.

eile:
ready for review.

kaabimg:
Checking each time that the spike vector is sorted is bad for performance. Fixing the doc is better. We can also add an assert (isSorted(spikes));

eile:
We already iterate over each spike in append(). Doing the check there probably costs nothing.

hernando:
Except that throwing after having written something is not exception safe. The check and the write should be independent so for big spikes sets than do not fit in cache there's obviously a cost. However I don't think were are going to write > 10^2 MBs in a single call even in multi-million cell simulations because it's biologically impossible that all fire at the same time during a reporting delta.

PR #136, opened 2017-02-21
Spec for new python compartment report API

eile:


doc/feature/pyCompartmentReport.md
```diff
@@ -0,0 +1,63 @@
...
+============
+
+## Motivation
+
+Redesign the current BBPSDK-inspired API for also allowing voltage traces.
+The returned data should change to be a matrix[frame, compartment] instead
+of the current vector[compartments].
+
+## API
+
```
hernando: I've realized that the current method for extracting the report metadata doesn't work for stream reports. We have to discuss what's the best alternative because I have only one use case in mind, which is the way the report is used in RTNeuron. In this case, the best solution is to add getEndTime() to the report view to get an updated timestamp everytime load is called, in the same way it works for the SpikeReport right now.
What I'm not so sure is how to deal with the updates. In the SpikeReportReader the user can request the data in a tiny time window known to exist, so load is non-blocking and the overhead of the update can be minimal (but it actually it depends on the implementation, in our case it's not the minimum possible because you still have to collect the data).
In the compartment report there's no method that will return en empty value immediately if the timestamps requested are out of the available window, the call will block instead. The only think you can do instead is to request data that you have already accessed, but I think it's very hard to avoid any internal copy of the frame data, which means that the API is enforcing an unavoidable performance penalty.

Note that with this approach the metadata from CompartmentReport should say that the time window is [0, inf) for stream reports. This makes them detectable and it's inconsistent with the SpikeReportReader, which lacks the view concept. Nonetheless, I'm realizing that maybe it makes sense to apply the view concept also to the SpikeReportReader, that way I could get rid of the wrapper that I have in RTNeuron to ensure thread-safety in case the user decides to use the report from the console at the same time than the engine. That would also solve the inconsistency nicely. It's just splitting the current API in two with one repeated functions (getEndTime).

We also need to add a function to know if the report is ended, this is absolutely necessary and I don't know where to put it. At the same time, we also need the close method to unlock any thread that may be blocked in load.

doc/feature/pyCompartmentReport.md
```diff
@@ -0,0 +1,63 @@
...
+    view = reader.create_view([1])
+    times, trace = view.load() # times: 1D double ndarray, trace: 2D float ndarray
+    for time, frame in zip(times, trace):
+        print "{}, {}".format(time, frame[soma_index(42)])
+
+    # full frame of sub-target
+    view = reader.create_view(range(1, 1000))
+    time, frame = view.load( 42.f ) # time: double, frame: 1D float ndarray
+    for gid in range(1, 1000):
+        print "{}, {}".format(gid, frame[soma_index(gid)])
```
hernando: Remove the return type comments from here.

hernando:
The null report shouldn't be called null but random.

eile:
ping

doc/feature/pyCompartmentReport.md
hernando: To summarize the comment above:
* The report metadata is static, that doesn't work from stream reports. A possible solution is:
    * Adding getEndTime to the view could be the solution...
    * but it's not clear when to update the timestamp.
    * The static metadata would remain and endtime would be +inf for stream reports.
    * This solution contains some inconsistencies compared to SpikeReportReader. Maybe the same principles should be applied there.
* Functions to now when a report is ended and close a stream (to unlock any reader) are missing.

PR #137, opened 2017-02-23
Dummy compartment report improvements

eile:


apps/compartmentConverter.cpp
```diff
@@ -61,6 +63,19 @@ template< class T > void requireEqualCollections( const T& a, const T& b )
...
 
+/** @return true if the cell occupies a continuous region in the report frame */
+bool _isContinuous( const brion::CompartmentReport& report,
+                    const size_t gidIndex )
+{
+    const auto& offsets = report.getOffsets()[gidIndex];
+    const auto& counts = report.getCompartmentCounts()[gidIndex];
+    for( size_t i = 0; i < offsets.size() - 1; ++i )
+        if( offsets[i] + counts[i] != offsets[i+1] )
+            return false;
```
hernando: Rename to _isContiguous or _isCompact

This is not completely accurate. A cell could be in a single contiguous region, but since the sections are not guaranteed to be enumerated in order inside the region, this function would return a false negative. It's more correct to do the following: let's call  [Si_a, Si_b] the offset range where a section Si maps (Si_a is the base offset and Si_b is Si_a + num compartments of that section), what we need to verify is that for any section Sx, [Sx_a, Si_b] is in [min(Si_a), max(Si_b)] and that max(Si_b) - min(Si_b) = num compartments for this cell. The second part is the key to ensure that all sections are compact.

brain/compartmentReportReader.h
```diff
@@ -216,8 +216,8 @@ class CompartmentReportMapping
 public:
     struct IndexEntry
     {
-        // To ensure proper alignment and compactness the order of this fields
-        // mustn't be changed.
+        // To ensure proper alignment and compactness for the python binding the
+        // order of this fields mustn't be changed.
```
hernando: type: this -> these

doc/feature/compartmentReportPy.md
```diff
@@ -0,0 +1,86 @@
+High-Level CompartmentReport API for Python {#compartment_report_py}
```
hernando: Hadn't you separated the PRs?

brion/plugin/compartmentReportDummy.cpp
```diff
@@ -0,0 +1,170 @@
+/* Copyright (c) 2015-2017, EPFL/Blue Brain Project
```
hernando: Could you add a test for the new feature?

apps/compartmentConverter.cpp
```diff
@@ -219,12 +233,23 @@ int main( const int argc, char** argv )
...
+                const float* values = &voltages[ offsets[index][0] ];
+                const size_t size = std::accumulate( counts[index].begin(),
+                                                     counts[index].end(), 0 );
+                if( !to.writeFrame( gid, values, size, t ))
+                    return EXIT_FAILURE;
+                ++index;
+                continue;
+            }
+
             brion::floats cellVoltages;
```
hernando: Remove the reference to voltage also here.

brion/plugin/compartmentReportDummy.cpp
```diff
@@ -0,0 +1,170 @@
...
+    _defaultGIDs = (i == uri.queryEnd( )) ? 1024 :
+                       boost::lexical_cast< size_t >( i->second );
+    _cacheNeuronCompartmentCounts( _gids );
+}
+
+std::string CompartmentReportDummy::getDescription()
+{
+    return R"(Benchmark drain (write): dummy://
+Benchmark dummy source (read): dummy://[?size=size-in-MB][&randomValues]
+    default 1024MB of pseudo-random report data)";
```
hernando: 1 GB instead of 1024MB

brion/plugin/compartmentReportDummy.cpp
```diff
@@ -0,0 +1,170 @@
...
+    floatsPtr buffer( new floats( getFrameSize( )));
+    if( _randomValues )
+    {
+        std::random_device dev;
+        std::mt19937_64 engine( dev( ));
+        std::uniform_int_distribution < int16_t > distribution;
+        engine.seed( unsigned( time )); // reproducible randomness, please
+
+        for( float& value : *buffer )
+            value = float( distribution( engine )) / 1000.f;
```
hernando: Maybe it would be good to have the floating point numbers generated in a user defined range. At least it should be documented which is the default range (which looks a bit strange I have to say).

apps/compartmentConverter.cpp
eile: Since this is an optimization, a false negative is not a concern. I'll see if I find an easy implementation for your comment.

apps/compartmentConverter.cpp
eile: done

brain/compartmentReportReader.h
eile: done (this was yours ;) )

doc/feature/compartmentReportPy.md
eile: Yes, but the other one is still not approved!

brion/plugin/compartmentReportDummy.cpp
eile: No, the size in in MB. I think one can do the conversion easily.

brion/plugin/compartmentReportDummy.cpp
eile: When there is a use case. It's useless random data, so kiss for now.

eile: NB: that's also why the range is not documented on purpose.

doc/feature/compartmentReportPy.md
hernando: But I was expecting this commit to be rebased out.

brion/plugin/compartmentReportDummy.cpp
```diff
@@ -0,0 +1,170 @@
...
+    std::mt19937_64 engine( dev( ));
+    std::uniform_int_distribution < uint32_t > distribution;
+    engine.seed( _defaultGIDs ); // reproducible randomness, please
+
+    GIDSet all;
+    while( all.size() < _defaultGIDs )
+        all.insert( distribution( engine ));
+
+    const auto& subset = gids.empty() ? all : gids;
+    _gids = _computeIntersection( all, subset );
```
hernando: Replace with:

    if (gids.empty())
        _gids = std::move(all);
    else
        _gids = _computeIntersection(all, gids)

brion/plugin/compartmentReportDummy.cpp
eile: done

PR #130, opened 2017-01-30
brain::CompartmentReportReader API

kaabimg:
brain::CompartmentReportReader API

PR #138, opened 2017-03-06
brain::CompartmentReport : python API v2

kaabimg:


brain/compartmentReportReader.cpp
```diff
@@ -178,6 +182,45 @@ std::future<CompartmentReportFrames> CompartmentReportView::load(float start,
...
+        double timestep = _impl->report->getTimestep();
+        auto loadFrameTask = [start, end, timestep, report, id] {
+
+            CompartmentReportFrame frame;
+
+            try
+            {
+                auto data = report->loadNeuron(id);
+                if (data)
+                    frame._impl->data = std::move(*data);
```
hernando: This doesn't look correct to me. I think that the only proper way of using loadNeuron without having to do an extra copy here (and another one in the wrapping) is to get rid of the ComparmentReportFrame class and return the data directly in a matrix-like object with the same layout expected by the numpy array. Here you're passing an array of float point values for all frames of a neuron into a single frame, which conceptually is wrong.
Note that the function returns different results for the single cell case depending on the implementation supporting loadNeuron or not and that shouldn't be the case.
For this PR I'd focus on the Python side and leave this for later because I don't see how to do it efficiently without changes in the C++ API.

brain/python/compartmentReportReader.cpp
```diff
@@ -173,24 +150,24 @@ bp::object CompartmentReportMapping_getCompartmentCounts(
     return result;
 }
 
-void export_CompartmentReportReader()
+void export_CompartmentReport()
 {
-// clang-format off
+    // clang-format off
```
hernando: do not indent

brain/python/numpyArrray.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
```
hernando: Why have you moved this to a separate file? I prefer to keep it inside the only .cpp that uses it.

tests/brain/python/compartmentReportReader.py
```diff
@@ -54,51 +55,80 @@ def test_mapping(self):
 
     def test_frames(self):
         view = self.report.create_view({1,2,3})
-        frames = view.load(0.0,0.2)
-        assert(len(frames) == 2)
-        frames = view.load(0.05,0.25)
+        timestamps ,frames = view.load(0.0,0.2)
```
hernando: timestamps, frames

review also below.

tests/brain/python/compartmentReportReader.py
```diff
@@ -29,15 +29,16 @@
...
         assert(metadata['data_unit'] == 'mV')
         assert(metadata['time_unit'] == 'ms')
         assert(metadata['start_time'] == 0.0)
         assert(metadata['end_time'] == 10.0)
-        assert(metadata['time_step'] == 0.10000000149011612)
+        assert(abs(metadata['time_step'] - 0.1) < 0.0000000149011612)
+
 
     def test_create_view(self):
         view = self.report.create_view({1,2,3})
```
hernando: Add one space after the comma. Review also below.

brain/compartmentReportReader.cpp
```diff
@@ -120,7 +124,7 @@ const CompartmentReportMapping& CompartmentReportView::getMapping() const
 }
 
 std::future<CompartmentReportFrame> CompartmentReportView::load(
-    const float timestamp)
+    double timestamp)
```
eile: should stay const

brain/compartmentReportReader.cpp
```diff
@@ -178,6 +182,45 @@ std::future<CompartmentReportFrames> CompartmentReportView::load(float start,
 
 std::future<CompartmentReportFrames> CompartmentReportView::loadAll()
 {
+    if (_impl->report->getGIDs().size() == 1)
```
eile: invert logic for early out:
```
if( size != 1)
  return load(...)
  
[rest of code]
```

tests/brain/python/compartmentReportReader.py
```diff
@@ -54,51 +55,80 @@ def test_mapping(self):
...
+
+        assert(timestamp == 0.1)
+        assert(frame.tolist() ==
                [-65.14350891113281, -65.29447937011719, -65.4448013305664])
 
     def test_frame_data(self):
-        data = CompartmentReportReader(report_path).create_view(
-            {1, 2, 3}).load(0.1).data
+        time, data = CompartmentReport(report_path).create_view(
+            {1, 2, 3}).load(0.1)
```
hernando: Verify also the timestamp

tests/brain/python/compartmentReportReader.py
```diff
@@ -54,51 +55,80 @@ def test_mapping(self):
...
 
-        mapping = CompartmentReportReader(report_path).create_view(
+        mapping = CompartmentReport(report_path).create_view(
             {1, 2, 3}).mapping
         assert(mapping.num_compartments(1) == 1)
         assert(mapping.index.tolist() ==
                [(0, 1, 0, 1), (1, 2, 0, 1), (2, 3, 0, 1)])
 
+
+class TestExample(unittest.TestCase):
```
hernando: I think there's no need to add a test with the example code from the spec. Just verify that existing tests have the proper code coverage or add a new one if needed.

brain/python/arrayHelpers.cpp
```diff
@@ -239,6 +239,52 @@ bp::object toNumpy( const brain::Matrix4f& matrix )
...
+
+bp::object frameToTuple(CompartmentReportFrame&& frame)
+{
+    return bp::make_tuple(frame.getTimestamp(), toNumpy(frame.takeData()));
+}
+
+bp::object framesToTuple(CompartmentReportFrames&& frames)
+{
+    if(frames.empty())
+        return bp::make_tuple(-1.f,std::vector<float>());
```
hernando: Return None instead of this.

brain/python/arrayHelpers.cpp
```diff
@@ -239,6 +239,52 @@ bp::object toNumpy( const brain::Matrix4f& matrix )
...
+    if(frames.size() == 1)
+        return frameToTuple(std::move(frames.front()));
```
hernando: This violates the API contract. Even if the list has a single element, this function has still to return a list of timestamps and and a 2D matrix. Otherwise the users have to check for this particular case whenever calling load(start, end) or loadAll(), otherwise their Python code may raise exceptions.

brain/python/compartmentReportReader.cpp
```diff
@@ -105,46 +99,29 @@ CompartmentReportMappingProxy CompartmentReportView_getMapping(
...
 }
 
-CompartmentReportFramePtr CompartmentReportView_loadAt(
-    CompartmentReportView& view, float time)
+
+bp::object CompartmentReportView_loadAt(CompartmentReportView& view, double time)
 {
-    return CompartmentReportFramePtr(
-        new CompartmentReportFrame(view.load(time).get()));
+    return frameToTuple(view.load(time).get());
```
hernando: I've noticed that the C++ implementation of this function has a bug. If passes a timestamp of the frame the requested timestamp when it should return the timestamp of the start of the time window of the returned frame. 
Furthermore, the frame could be empty, which happens when the requested timestamp is outside the report boundaries. In this case you have to check frame.empty() and return None if that's the case.

brain/compartmentReportReader.h
```diff
@@ -13,6 +13,7 @@
  * FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
  * details.
  *
+ *
  * You should have received a copy of the GNU Lesser General Public License
  * along with this library; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
```
hernando: Split this file in one file per class. You may keep CompartmentReportMetadata with CompartmentReport.

brain/python/arrayHelpers.cpp
```diff
@@ -239,6 +239,52 @@ bp::object toNumpy( const brain::Matrix4f& matrix )
...
+
+    std::vector<double> timestamps;
+    timestamps.reserve(frames.size());
+    for (const auto& frame : frames)
+        timestamps.push_back(frame.getTimestamp());
+
+    size_t size = frames.size() * frames[0].getData().size();
```
hernando: assign frames[0].getData().size() to a variable to reuse it in the sentence below.

brain/python/arrayHelpers.cpp
```diff
@@ -239,6 +239,52 @@ bp::object toNumpy( const brain::Matrix4f& matrix )
...
+
+    for (const auto& frame : frames)
+    {
+        std::copy(frame.getData().begin(), frame.getData().end(), matrixData);
+        matrixData += frame.getData().size();
+    }
+
+    PyObject* array =
+        PyArray_New(&PyArray_Type, 2, dimensions, NPY_FLOAT, 0, matrixData, 0,
+                    NPY_ARRAY_OWNDATA | NPY_ARRAY_F_CONTIGUOUS, 0);
```
hernando: I think the layout must be NPY_ARRAY_C_CONTIGUOUS. Furthermore, in the tests you check the shape of the ouput numpy array, but there's no test for the values themselves (except for the non matrix case).
I even think you can use PyArray_SimpleNewFromData, which has fewer parameters.

brain/compartmentReportReader.cpp
kaabimg: This males the code inconsistent with load(double start, double end) and with the method declaration signature (which also makes my code editor not working properly).

brain/compartmentReportReader.cpp
kaabimg: Ok

brain/compartmentReportReader.cpp
```diff
@@ -178,6 +182,45 @@ std::future<CompartmentReportFrames> CompartmentReportView::load(float start,
...
+    {
+        auto report = _impl->report;
+        auto id = *_impl->report->getGIDs().begin();
+
+        double start = _impl->report->getStartTime();
+        double end = _impl->report->getEndTime();
+        double timestep = _impl->report->getTimestep();
```
kaabimg: Ok

brain/compartmentReportReader.cpp
kaabimg: Indeed, the frame in this case is containing a trace. As you mentioned, this would be changed later.

brain/python/compartmentReportReader.cpp
kaabimg: Ok

brain/python/numpyArrray.h
kaabimg: When I included the numpy header (before moving the code to arrayHelpers), I got a bunch of warnings that you disabled by some defines. That is why I moved all that in a single header which is only included from arrayHelpers.cpp but that can be reused later.

tests/brain/python/compartmentReportReader.py
kaabimg: Ok

tests/brain/python/compartmentReportReader.py
kaabimg: Ok

brain/python/arrayHelpers.cpp
kaabimg: Ok

brain/python/arrayHelpers.cpp
kaabimg: This is not what I understood from the spec. Indeed, I found it strange that, depending on the frames size, it changed the layout of the returned value, but : 
 def load(timestamp):
......
The return type is a (double, nadarray) tuple where the first element is the
        actual timestamp of the frame and the second one is a 1D double ndarray."""

brain/compartmentReportReader.cpp
kaabimg: Ok

brain/python/compartmentReportReader.cpp
kaabimg: Can't do it. The pre-commit hook will reinsert the indent

tests/brain/python/compartmentReportReader.py
kaabimg: This test is duplicated, will be removed. The other one contains all the checks

brain/compartmentReportView.cpp
```diff
@@ -120,9 +56,23 @@ const CompartmentReportMapping& CompartmentReportView::getMapping() const
 }
 
 std::future<CompartmentReportFrame> CompartmentReportView::load(
-    const float timestamp)
+    double timestamp)
 {
+    const double start = _impl->report->getStartTime();
+    const double end = _impl->report->getEndTime();
+
+    if (timestamp < start || timestamp > end)
```
hernando: if timestamp == end it should also be considered invalid because the frame time window interval is open on the right.

brain/compartmentReportView.cpp
```diff
@@ -120,9 +56,23 @@ const CompartmentReportMapping& CompartmentReportView::getMapping() const
...
+        throw std::logic_error("Invalid timestamp");
+
+    const double timestep = _impl->report->getTimestep();
+
+    size_t frameIndex = std::min(size_t(round((end - start) / timestep)) - 1,
+                                 size_t(round(timestamp / timestep)));
+
+    timestamp = start + frameIndex * timestep;
```
hernando: You already know timestamp is in [start, end) so replace lines 69-72 with:
timestamp = start + std::floor((timestamp - start) / timestep) * timestep;

brain/compartmentReportView.cpp
```diff
@@ -163,7 +113,7 @@ std::future<CompartmentReportFrames> CompartmentReportView::load(float start,
         for (size_t i = std::floor(start / timestep); i * timestep < end; ++i)
```
hernando: The code here is now exactly the same as in load(timestamp) because here it's assumed that the report starts at a time which is multiple of the timestep. I don't know if it could not be the case.

tests/brain/python/compartmentReport.py
```diff
@@ -54,50 +54,66 @@ def test_mapping(self):
...
     def test_frames(self):
         view = self.report.create_view({1,2,3})
-        frames = view.load(0.0,0.2)
-        assert(len(frames) == 2)
-        frames = view.load(0.05,0.25)
+        timestamps, frames = view.load(0.0,0.2)
+
+        assert(numpy.isclose(timestamps,[ 0.0, 0.1]).all())
+        assert(numpy.isclose(frames, [[1.58346726e-43, 0.0e+0, 4.56420264e-37],
+            [0.0e+0, 0.0e+0, 0.0e+0]]).all())
```
hernando: Are you sure about this?, these numbers don't look right.

tests/brain/python/compartmentReport.py
```diff
@@ -54,50 +54,66 @@ def test_mapping(self):
...
-        assert(frame.timestamp == 0.10000000149011612)
-        assert(frame.empty == False)
-        assert(frame.data.tolist() ==
-               [-65.14350891113281, -65.29447937011719, -65.4448013305664])
 
-    def test_frame_data(self):
-        data = CompartmentReportReader(report_path).create_view(
-            {1, 2, 3}).load(0.1).data
-        assert(data.tolist() ==
-               [-65.14350891113281, -65.29447937011719, -65.4448013305664])
```
hernando: Add a test for loading a range from a temporary view.

brain/compartmentReportView.cpp
```diff
@@ -120,9 +56,22 @@ const CompartmentReportMapping& CompartmentReportView::getMapping() const
...
 {
+    const double start = _impl->report->getStartTime();
+    const double end = _impl->report->getEndTime();
+
+    if (timestamp < start || timestamp >= end)
+        throw std::logic_error("Invalid timestamp");
+
+    const double timestep = _impl->report->getTimestep();
+
+    size_t frameIndex = std::llround((timestamp - start) / timestep);
```
kaabimg: Actually, some tests are failing when using floor. 
I also would prefer to use the same thing as the low level CR (I hesitated to use any ugly extern before duplicating the code).
Either we change both or nothing.

hernando: Change both then

kaabimg: The tests failures that I get are real bugs. The problem is that brion::CR uses single precision floats.
floor doesn't work properly with single precision floats:
       float t = 9.4, timestep = 0.1;
       std::cout << "round " << (size_t)std::round(t / timestep) << std::endl;
       std::cout << "floor " << (size_t)std::floor(t / timestep) << std::endl;
>> round 94
>> floor 93
        double t = 9.4, timestep = 0.1;
        std::cout << "round " << (size_t)std::round(t / timestep) << std::endl;
        std::cout << "floor " << (size_t)std::floor(t / timestep) << std::endl;
>> round 94
>> floor 94

hernando: Move all timestamps in the low level part to double then?

hernando: On the other hand this shows that conceptually, while using timestamps is the right thing, maybe the actual type should be an integer type + units like chrono instead of a floating point number.

kaabimg: It would be fine for the binary and HDF5 plugins but will break the Map plugin since that it serializes a pod header with floats. If it is fine to break the compatibility of this plugin, yes we can move to doubles

hernando: Fine with me, but this should be checked with @eile.
At the moment all caching is temporary and the compartment data caching is not used in production.

doc/Changelog.md
```diff
@@ -2,7 +2,8 @@ Changelog {#Changelog}
 =========
 
 # git master
-
+* [138](https://github.com/BlueBrain/Brion/pull/138):
+  brain::CompartmentReport : python API v2
```
hernando: This changelog registers relevant changes between releases. The intra-release history is not relevant. That means that I'd remove this Changelog entry, pick 130, move it to the top and rewrite it as:

    * [130](https://github.com/BlueBrain/Brion/pull/130), [138](https://github.com/BlueBrain/Brion/pull/138):
      Added brain::CompartmentReport and its Python wrapping.

tests/brain/python/compartmentReport.py
```diff
@@ -0,0 +1,140 @@
...
+        assert(frames.shape == (3,3))
+        timestamps, frames = view.load(0.09,0.12)
+        # Despite the time window is smaller than the timestep, each edge is
+        # falling on a different frame, so we get two frames.
+        assert(len(timestamps) == 2)
+        assert(frames.shape == (2, 3))
+        timestamps, frames = view.load_all()
+        assert(len(timestamps) == 100)
+        timestamp, frame = view.load(0.1)
+        assert(numpy.isclose(timestamp, 0.1) )
```
hernando: no space between parentheses (clang-format won't reformat this)

brain/compartmentReportView.cpp
```diff
@@ -119,10 +55,30 @@ const CompartmentReportMapping& CompartmentReportView::getMapping() const
     return _impl->mapping;
 }
 
+// using float is necessary to match brion::CompartmentReport
+// frame index calculation
+inline size_t getFrameIndex(float t, float start, float timestep)
```
hernando: Why float and not double here?

hernando: Put inside an anonymous namespace and prepend a _ to the name to stress it's a private function.

brain/compartmentReportView.cpp
```diff
@@ -119,10 +55,30 @@ const CompartmentReportMapping& CompartmentReportView::getMapping() const
...
+    const double end = _impl->report->getEndTime();
+
+    if (timestamp < start || timestamp >= end)
+        throw std::logic_error("Invalid timestamp");
+
+    const double timestep = _impl->report->getTimestep();
+
+    size_t frameIndex = getFrameIndex(timestamp, start, timestep);
+
+    timestamp = start + frameIndex * timestep;
```
hernando: Since you only use getFrameIndex here, you could move this sentence to it and rename it to _snapTimestamp, _snapToTimeResolution or something similar

PR #142, opened 2017-03-17
Update README

eile:


README.md
```diff
@@ -1,6 +1,4 @@
-[TOC]
-
-# Introduction {#Introduction}
+# Brion
```
hernando: Have you verified how does this look in Doxygen? I think that with your change the project name appears twice.

README.md
```diff
@@ -59,19 +57,19 @@ The higher level library is called Brain and it provides:
...
 
 Building from source is as simple as:
 
     git clone https://github.com/BlueBrain/Brion.git
     mkdir Brion/build
     cd Brion/build
-    cmake ..
-    make
+    cmake -GNinja ..
+    ninja
```
hernando: Ninja is not known by everybody, I'd keep it as it was.

eile:
Both of these changes should have been done and discussed during the last release.

hernando:
If all the rest are the same, then OK

PR #145, opened 2017-04-04
CompartmentReport : async io

kaabimg:


kaabimg:
Retest this please.

kaabimg:
Retest this please

eile:
Please separate the clang-format into a separate commit

kaabimg:
Done, except for compartmentConverter.cpp which doesn't pass if not properly formatted.

brain/compartmentReportView.cpp
```diff
@@ -95,45 +95,18 @@ std::future<CompartmentReportFrame> CompartmentReportView::load(
     return _impl->readerImpl->threadPool.post(loadFrameTask);
```
hernando: Wasn't it the idea to get rid of CompartmentReportFrame completely and use a simple double floatsPtr struct instead (as Frames for multiple vectors)? The current API in CompartmentReportView looks a bit asymmetrical otherwise. Here you can use a deferred async to transfer the output from the brion::CompartmentReport::loadFrame to the struct.

brion/compartmentReport.cpp
```diff
@@ -122,12 +124,27 @@ size_t CompartmentReport::getBufferSize() const
...
 
-floatsPtr CompartmentReport::loadFrame(const float timestamp) const
+std::future<floatsPtr> CompartmentReport::loadFrame(const float timestamp) const
+
 {
-    if (timestamp < getStartTime() || timestamp > getEndTime())
-        return floatsPtr();
+    auto task = [timestamp, this] {
+        if (timestamp < getStartTime() || timestamp > getEndTime())
+            return floatsPtr();
```
hernando: As said in the app, throw instead of returning an empty array?

brion/compartmentReport.h
```diff
@@ -115,41 +118,45 @@ class CompartmentReport : public boost::noncopyable
...
      * @return the report values if found at timestamp, nullptr otherwise
-     * @version 1.0
+     * @note Until the completion of this operation, the report shouldn't be
+     * modified
+     * @version 2.0
      */
-    BRION_API floatsPtr loadFrame(float timestamp) const;
+    BRION_API std::future<floatsPtr> loadFrame(float timestamp) const;
+
+    BRION_API std::future<Frames> loadFrames(float start, float end) const;
```
hernando: Add documentation. For example, it's not obvious that if the requested window is not contained in the report window, this functions returns an empty frame list instead.
On the other hand, I'm not sure that I agree with that behaviour. If you don't change the functions to throw instead of returning an empty array, I think that loadFrames should load as many frames as possible within the start, end input interval (regardless of it being larger the report time window).

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -270,12 +265,31 @@ floatsPtr CompartmentReportHDF5::loadFrame(const float timestamp) const
...
+    return true;
+}
+
+Frames CompartmentReportHDF5::loadFrames(const float start,
+                                         const float end) const
+{
+    if (_offsets.empty())
+        return Frames();
+    lunchbox::ScopedWrite mutex(detail::_hdf5Lock);
+    return CompartmentReportCommon::loadFrames(start, end);
```
hernando: What's wrong with keeping the hdf5 lock inside _loadFrame? As a matter of fact I'm quite sure you should keep it there. H5 is a crappy when it comes to multithreading. You have to options, either you compile it with "multithreading" support, which means every operations tries to lock a single global mutex, or you do it yourself.
If loadFrames could process multiple frames in parallel, this code would crash because every thread already holds the (external) lock and all will try to use HDF5 at the same time, which is a recipe for disaster (don't ask me why, but even when accessing different files, there are global data structures used by all of them).

brion/plugin/compartmentReportBinary.cpp
```diff
@@ -251,16 +348,101 @@ floatsPtr CompartmentReportBinary::loadFrame(const float timestamp) const
...
     {
 #pragma omp parallel for
         for (ssize_t i = 0; i < ssize_t(_subNumCompartments); ++i)
-            lunchbox::byteswap((*buffer)[i]);
+            lunchbox::byteswap(buffer[i]);
+    }
+    return true;
+}
+
+bool CompartmentReportBinary::_loadFrameAIO(const float timestamp,
```
hernando: I still need to review this function. The rest is reviewed.

brion/plugin/compartmentReportBinary.cpp
```diff
@@ -251,16 +348,101 @@ floatsPtr CompartmentReportBinary::loadFrame(const float timestamp) const
...
+    {
+        // clang-format off
+        auto cb = readAsync(
+            {
+                _fileDescriptor,
+                buffer,
+                _header.numCompartments * sizeof(float),
+                frameOffset
+            }
+        );
```
hernando: Wouldn't it be faster if you split the read in blocks of, let's say 1MB? That gives more opportunities to exploit the parallelism of GPFS, right?

brion/plugin/compartmentReportBinary.cpp
```diff
@@ -251,16 +348,101 @@ floatsPtr CompartmentReportBinary::loadFrame(const float timestamp) const
...
+    if (!_subtarget)
+    {
+        // clang-format off
+        auto cb = readAsync(
+            {
+                _fileDescriptor,
+                buffer,
+                _header.numCompartments * sizeof(float),
+                frameOffset
+            }
```
hernando: This block confused me a little bit at first sight, I thought it was a lambda. I prefer initializer lists to be styled the same as function calls.

brion/plugin/compartmentReportBinary.cpp
```diff
@@ -251,16 +348,101 @@ floatsPtr CompartmentReportBinary::loadFrame(const float timestamp) const
...
+                   _fileDescriptor,
+                    buffer + targetOffset,   // destination buffer
+                    numCompartments * sizeof(float), // size
+                    frameOffset + sourceOffset * sizeof(float) //  global offset
+                }
+            );
+            // clang-format on
+        }
     }
+
```
hernando: Try to refactor the code to not repeat the lines 428-438. If you use the same aio code for both single and multiple operations it would be simple, for example you can use a different helper function to create the std::vector<AIOReadData> depending on the case. If the vector is empty you can return false (that would be the case _subNumCompatments == 0)

brion/plugin/compartmentReportBinary.cpp
```diff
@@ -251,16 +348,101 @@ floatsPtr CompartmentReportBinary::loadFrame(const float timestamp) const
...
+            const uint64_t targetOffset = offsets[i][j];
+
+            // clang-format off
+            readData.push_back(                      
+                {
+                   _fileDescriptor,
+                    buffer + targetOffset,   // destination buffer
+                    numCompartments * sizeof(float), // size
+                    frameOffset + sourceOffset * sizeof(float) //  global offset
+                }
```
hernando: I'd try to put all the cells that are contiguous in the same read operation until a maximum size is reached, it shouldn't be very difficult to do. The question is, do you think that can improve performance?

hernando:
I've pushed my review but there are still some comments that we need to decide whether we want to address them now or in a separate pull request.

hernando:
The changelog still needs updating also.

apps/compartmentConverter.cpp
```diff
@@ -234,17 +234,37 @@ int main(const int argc, char** argv)
...
+            LBERROR << std::endl
+                    << "Can't load frame at " << t << " ms" << std::endl;
+            ::exit(EXIT_FAILURE);
+        }
         loadTime += clock.getTimef();
         if (!data)
         {
-            LBERROR << "Can't load frame at " << t << " ms" << std::endl;
+            LBERROR << std::endl
+                    << "Can't load frame at " << t << " ms" << std::endl;
```
kaabimg: This can go to another pull request where the error handling will be done through exceptions

brain/python/arrayHelpers.cpp
```diff
@@ -245,41 +245,38 @@ bp::object frameToTuple(CompartmentReportFrame&& frame)
...
-    size_t frameCount = frames.size();
-    size_t frameSize = frames[0].getData().size();
+    size_t frameCount = frames.timeStamps->size();
+    size_t frameSize = frames.data->size() / frameCount;
 
-    std::vector<double> timestamps;
-    timestamps.reserve(frameCount);
-    for (const auto& frame : frames)
-        timestamps.push_back(frame.getTimestamp());
+    // TODO :remove this when time stamps will be doubles in brion::CR
```
kaabimg: Ok

tests/brain/compartmentReport.cpp
```diff
@@ -223,19 +223,19 @@ void testReadRange(const char* relativePath)
...
-    BOOST_CHECK_EQUAL(frames[1].getTimestamp(), start + step);
+    BOOST_REQUIRE_EQUAL(frames.timeStamps->size(), 2);
+    BOOST_CHECK_EQUAL((*frames.timeStamps)[0], start);
+    BOOST_CHECK_EQUAL((*frames.timeStamps)[1], start + step);
 
     frames = view.load(start + 0.5 * step, start + step * 2.5).get();
-    BOOST_REQUIRE_EQUAL(frames.size(), 3);
-    BOOST_CHECK_EQUAL(frames[0].getTimestamp(), start);
-    BOOST_CHECK_EQUAL(frames[1].getTimestamp(), start + step);
-    BOOST_CHECK_EQUAL(frames[1].getTimestamp(), start + 2 * step);
```
kaabimg: Yes, just replaced by:
BOOST_CHECK_EQUAL((*frames.timeStamps)[0], start);
....
since that the type of frames changed

tests/compartmentReport.cpp
```diff
@@ -124,46 +124,35 @@ BOOST_AUTO_TEST_CASE(test_create_write_report)
                       std::runtime_error);
 }
 
-BOOST_AUTO_TEST_CASE(test_bounds_binary)
+void testBounds(const char* relativePath)
```
kaabimg: Ok

brion/types.h
```diff
@@ -110,6 +110,12 @@ typedef std::vector<Spike> Spikes;
 /** A list of Spikes events per cell gid, indexed by spikes times. */
 typedef std::multimap<float, uint32_t> SpikeMap;
 
+struct Frames
```
kaabimg: Ok

brain/compartmentReportView.cpp
kaabimg: This can be done in a separate pull request. This one was just to make the API asynchronous.

hernando:
Retest this please

brion/types.h
hernando: OK... but you haven't done it yet. I'm finishing it.

PR #154, opened 2017-05-24
Release preparations

hernando:


CMakeLists.txt
```diff
@@ -10,6 +10,9 @@ set(Brion_VERSION_ABI 9)
 
 list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/CMake
   ${CMAKE_SOURCE_DIR}/CMake/common ${CMAKE_SOURCE_DIR}/Pydoxine/CMake)
+if(NOT EXISTS ${CMAKE_SOURCE_DIR}/CMake/common/Common.cmake)
+  message(FATAL_ERROR "CMake/common missing, run: git submodule update --init")
+endif()
 
 include(GitExternal)
```
tribal-tec: not needed anymore

hernando: Actually it is, but Pydoxine. Unless we make this also a submodule, but then it'd be replicated by RTNeuron as we discussed.

tribal-tec: Ah, sure, my bad.

tribal-tec: But CMake/common is not removed yet from .gitexternals?

hernando: That I forget to do it.

PR #157, opened 2017-05-30
Improve synapse attribute reading performance, add Synapse.gid() python wrapping

tribal-tec:


brion/synapse.cpp
```diff
@@ -223,14 +221,25 @@ class Synapse : public boost::noncopyable
...
+            // OPT: check if we have at least one unmerged file before fetching
+            // all the filenames. Much faster in case there is no merged and/or
+            // unmerged file at all.
+            if (!boost::filesystem::exists(fs::path(source + ".0")))
+            {
+                throw std::runtime_error(
+                    "No merged or unmerged synapse file found: " + source);
+            }
+
+            LBWARN << "Found unmerged synapse files for " << source
```
hernando: Only unmerged synapses file found for

tribal-tec:
It throws an exception which is not catched by brain::Circuit (in contrast to nrn_extra), so the user gets it.

doc/Changelog.md
```diff
@@ -1,8 +1,12 @@
 Changelog {#Changelog}
 =========
 
-# Release 2.0.0 (24-May-2017)
+# Release 2.0.0 (30-May-2017)
 
+* [157](https://github.com/BlueBrain/Brion/pull/157):
+  Add Synapse.gid() python wrapping
+* [157](https://github.com/BlueBrain/Brion/pull/157):
```
hernando: repeated 157? In this chagelog we use bullet points for PRs with multiple changes. I've never seen the other style in any other project.

PR #161, opened 2017-06-15
Lunchbox no longer leaks out of Brion/Brain; doxygen fixes

tribal-tec:


brain/neuron/types.h
```diff
@@ -42,7 +42,7 @@ enum class SectionType
     undefined = brion::enums::SECTION_UNDEFINED
 };
 
-typedef boost::shared_ptr<Morphology> MorphologyPtr;
+typedef std::shared_ptr<Morphology> MorphologyPtr;
```
hernando: std::shared_ptr doesn't behave the same inside Python wrappings, as a matter of fact I think I will revert that change in RTNeuron because it may be leaking. Could you run the morphology tests under valgrind to see if that's the case.

PR #162, opened 2017-06-16
Add Pydoxine as a git submodule to ease nix build

tribal-tec:


tribal-tec:
@hernando FYI

tribal-tec:
Also needs https://github.com/BlueBrain/Pydoxine/pull/10 to fix building with nix

PR #165, opened 2017-06-28
Final tweaks for Jenkins wheel deployment

tribal-tec:
* build docker image with CMake, h5, boost, doxygen to speedup build
* necessary changes for BBP Jenkins VM infrastructure

### DIFF
--- a/packaging/python_wheel/Dockerfile
+++ b/packaging/python_wheel/Dockerfile
@@ -1,0 +1,45 @@
+FROM quay.io/pypa/manylinux1_x86_64
+RUN yum install -y zlib-devel bzip2-devel flex
+
+############
+# CMake
+############
+
+# Need to build this from source, since the binaries from Kitware don't
+# run with the old version of the libraries in this Centos version
+RUN wget --no-check-certificate http://cmake.org/files/v3.7/cmake-3.7.2.tar.gz && tar xfz cmake-3.7.2.tar.gz && cd cmake-3.7.2 && ./configure && make -j2 install
+RUN rm -rf cmake-3.7.2.tar.gz cmake-3.7.2
+
+
+############
+# HDF5
+############
+
+RUN wget --no-check-certificate http://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.8.17/src/hdf5-1.8.17.tar.bz2 && tar xfj hdf5-1.8.17.tar.bz2
+RUN cd hdf5-1.8.17 && ./configure CFLAGS=-fPIC CXXFLAGS=-fPIC --prefix=/usr/local --enable-cxx --with-pic --disable-shared --with-zlib=/usr/include/,/usr/lib64/ && make -j2 install
+
+# --disable-shared doesn't add libz to the archive, so it
+# must be done manually at the end, there aren't any collisions in
+# the .a file, so currently safe
+RUN cd hdf5-1.8.17 && ar x /usr/lib64/libz.a && ar r /usr/local/lib/libhdf5.a *.o
+RUN rm -rf hdf5-1.8.17.tar.bz2 hdf5-1.8.17
+
+
+############
+# Boost
+############
+
+RUN wget --no-check-certificate http://downloads.sourceforge.net/project/boost/boost/1.59.0/boost_1_59_0.tar.gz && tar xfz boost_1_59_0.tar.gz
+RUN cd boost_1_59_0 && ./bootstrap.sh --with-libraries=date_time,filesystem,program_options,regex,serialization,system,test && ./b2 -j2 -q cxxflags=-fPIC cflags=-fPIC threading=multi link=static --build-type=minimal install
+RUN cd boost_1_59_0 && ./b2 --clean
+RUN rm -rf boost_1_59_0.tar.gz # don't delete build folder, needed to build boost.python for different python versions on-the-fly
+
+############
+# Doxygen
+############
+
+# Needed for docstrings generation and system doxygen is too old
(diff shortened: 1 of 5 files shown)


packaging/python_wheel/Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM quay.io/pypa/manylinux1_x86_64
```
mgeplf: sigh, this is the way I had it first, but switched over to having only to pull quay.io/pypa/manylinux1_x86_64

Probably better to do it this way.  I would be nice if we could pull this dockerfile out, and use it for all the builds.  I guess that's for the future.

packaging/python_wheel/build_docker_image.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
mgeplf: when is this run?

packaging/python_wheel/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+# run with the old version of the libraries in this Centos version
+RUN wget --no-check-certificate http://cmake.org/files/v3.7/cmake-3.7.2.tar.gz && tar xfz cmake-3.7.2.tar.gz && cd cmake-3.7.2 && ./configure && make -j2 install
+RUN rm -rf cmake-3.7.2.tar.gz cmake-3.7.2
+
+
+############
+# HDF5
+############
+
+RUN wget --no-check-certificate http://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.8.17/src/hdf5-1.8.17.tar.bz2 && tar xfj hdf5-1.8.17.tar.bz2
```
mgeplf: weird, I can't seem to get this file.  Also, --no-check-certificate isn't needed for http

packaging/python_wheel/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+# the .a file, so currently safe
+RUN cd hdf5-1.8.17 && ar x /usr/lib64/libz.a && ar r /usr/local/lib/libhdf5.a *.o
+RUN rm -rf hdf5-1.8.17.tar.bz2 hdf5-1.8.17
+
+
+############
+# Boost
+############
+
+RUN wget --no-check-certificate http://downloads.sourceforge.net/project/boost/boost/1.59.0/boost_1_59_0.tar.gz && tar xfz boost_1_59_0.tar.gz
```
mgeplf: can be https, just tested w/o using `--no-check-certificate`

packaging/python_wheel/docker_build_wheel.sh
```diff
@@ -102,37 +44,39 @@ build_brain()
...
-build_doxygen
+    cd ..
+    rm -rf build
+}
 
 for version in $PYTHON_VERSIONS; do
     build_brain $version
 done
+
+chown -R $UID.$UID $WHEELHOUSE
```
mgeplf: setting the group to the UID isn't needed, imo

packaging/python_wheel/docker_build_wheel.sh
```diff
@@ -102,37 +44,39 @@ build_brain()
...
     auditwheel repair -w $WHEELHOUSE dist/*${version}*
     popd
-}
 
-install_reqs
-build_cmake
-build_hdf
-build_doxygen
+    cd ..
+    rm -rf build
```
mgeplf: you don't want to keep the build files around?

packaging/python_wheel/Dockerfile
tribal-tec: Agreed. It's a start for now.

packaging/python_wheel/Dockerfile
tribal-tec: Works on the VM which is different than your local machine. I'd also keep --no-check-certificate just in case the proxy does weird things.

packaging/python_wheel/Dockerfile
tribal-tec: does not work on the VM with proxy, trust me :)

packaging/python_wheel/build_docker_image.sh
tribal-tec: conditionally as part of the Jenkins plan. The build has a parameter the user can switch to 'on', 'off' by default.

packaging/python_wheel/docker_build_wheel.sh
tribal-tec: no. we build 3 times in that folder, so better be clean everytime.

packaging/python_wheel/docker_build_wheel.sh
tribal-tec: got this from the pybin reports plan. You should better fix it there as well :)

tribal-tec:
@hernando any comments? I would need you for approval to merge this.

packaging/python_wheel/Dockerfile
mgeplf: wat

packaging/python_wheel/Dockerfile
mgeplf: sigh

packaging/python_wheel/build_docker_image.sh
mgeplf: Ah, good idea.  Maybe I'll do the same.

packaging/python_wheel/docker_build_wheel.sh
mgeplf: Ok, works for me.

packaging/python_wheel/docker_build_wheel.sh
mgeplf: Yup.

PR #166, opened 2017-07-05
VIZTM-732 : brain.CompartmentReport improvements

kaabimg:


brain/compartmentReportView.h
```diff
@@ -78,6 +78,22 @@ class CompartmentReportView
      */
     BRAIN_API std::future<brion::Frames> load(double start, double end);
 
+    /** Load frames between start and end time stamps.
+     *
+     * @param start the start time stamp
+     * @param end the end time stamp
+     * @param stride the time stride
```
hernando: @arsenius7, Does the stride need to be a timestep or the number to frame to be advanced?

To me using time for the stride is bogus. What sense does it have to use a stride of timestep * 1.1 if there's no frame interpolation? That simply means that the frame skip is a bit unpredictable.

brain/compartmentReportView.cpp
```diff
@@ -99,6 +99,36 @@ std::future<brion::Frames> CompartmentReportView::load(double start, double end)
...
+
+    start = std::max(start, _impl->report->getStartTime());
+    end = std::min(end, _impl->report->getEndTime());
+
+    auto task = [this, start, end, stride] {
+
+        brion::Frames frames;
+        frames.timeStamps.reset(new brion::doubles);
+        frames.data.reset(new floats);
+        for (double t = start; t < end; t += stride)
```
hernando: This loop is troublesome expressed this way, there are going to be cases of rounding errors.
In fact using a floating point stride is troublesome in itself.
It's better to write a while loop that increases a counter, the condition will be start + i * stride < end.
With stride being a integer with the number of frames to advance, if would be start + timestep*0.5 + i * stride * timestep < end

brain/compartmentReportView.cpp
```diff
@@ -99,6 +99,36 @@ std::future<brion::Frames> CompartmentReportView::load(double start, double end)
     return _impl->report->loadFrames(start, end);
 }
 
+std::future<brion::Frames> CompartmentReportView::load(double start, double end,
```
hernando: Do you have an estimate of the performance comparison between loadAll and load all with stride 2*timestep?

brain/python/compartmentReport.cpp
```diff
@@ -90,6 +90,10 @@ bp::object CompartmentReport_getMetaData(const CompartmentReport& reader)
     dict["time_step"] = md.timeStep;
     dict["time_unit"] = md.timeUnit;
     dict["data_unit"] = md.dataUnit;
+    dict["num_cells"] = md.cellCount;
+    dict["num_compartments"] = md.compartmentCount;
+    dict["num_frames"] = md.frameCount;
```
kaabimg: In C++, the convention is to use count.
But the request was "num_*". May be a python convention. I'm not sure

brain/compartmentReportView.h
```diff
@@ -78,6 +78,22 @@ class CompartmentReportView
      */
     BRAIN_API std::future<brion::Frames> load(double start, double end);
 
+    /** Load frames between start and end time stamps.
```
kaabimg: Ok

brain/compartmentReportView.h
```diff
@@ -78,6 +78,22 @@ class CompartmentReportView
...
+     *
+     * @param start the start time stamp
+     * @param end the end time stamp
+     * @param stride the time stride
+     * @return the frames overlapped by the given time window, spaced by a given
+     * stride. The start time
+     *         doesn't need to be aligned with the stride and the time
+     *         interval is open on the right. The result may be empty if the
+     *         time window falls out of the report window.
```
kaabimg: Ok

brain/python/compartmentReport.cpp
arsenius7: I don't really care.
If you prefer 'count',  let it be 'count'.

brain/compartmentReportView.h
arsenius7: First of all, imho, stride should be in the same units as start and end.
So if start and end are times in milliseconds, I would expect stride to be the same.

Restricting it to be a multiple of time_step should suffice; thus no need for interpolation.

arsenius7: Again, if you have _strong_  preference for number of frames, I could control it from BluePy side and pass number of frames here.

arsenius7: In my case I converted time_start, time_end and time_step to integers prior to passing it to C++ reader, so that one was working with integers only.

arsenius7: It's not like I'm suggesting to do it here, of course. :)

hernando: If "stride" is not going to mean the same as stride in an array, then it should be called step. To me it would be confusing otherwise.

hernando: I don't care very much a bout num or count, but I'd like it to be the same in both C++ and Python (except for camelCase versus underscores). For other parts of the wrapping it's not a big issue, but this one in particular changes quite a lot. I wouldn't mind the names different if the Python names are documented in CompartmentReport::getMetaData (and only for the Python docs), actually I think this should be done anyway. @kaabimg let me know if it's not clear what I mean.

arsenius7: I trust your judgement wrt to names, so name it here whatever you like (accidentaly, BluePy exposes this parameter as 'step'; in my request I named it 'stride' for the reason I stated above: the previous C++ layer was accepting frame numbers and not times).

brain/compartmentReportView.cpp
```diff
@@ -99,6 +99,54 @@ std::future<brion::Frames> CompartmentReportView::load(double start, double end)
...
+    const double reportStartTime = _impl->report->getStartTime();
+
+    if (end <= start)
+        throw std::logic_error("Invalid interval");
+
+    if (step < reportTimeStep || step <= 0.)
+        throw std::logic_error("Invalid step");
+
+    // check step is multiple of timestep
+    if (fmod(step, reportTimeStep) != 0.0)
```
hernando: isn't != 0 a bit strict? Shouldn't it be > epsilon?

brain/compartmentReportView.cpp
```diff
@@ -99,6 +99,54 @@ std::future<brion::Frames> CompartmentReportView::load(double start, double end)
...
+        throw std::logic_error(
+            "Step should be a multiple of the report time step");
+
+    /// check with Juan
```
hernando: Comment: Making sure the timestamps we are going to request always fall in the middle of a frame. For that we snap start to the beginning of the frame it's contained and then we add half the time step.

brain/compartmentReportView.h
```diff
@@ -78,6 +78,21 @@ class CompartmentReportView
...
+    /** Load frames between start and end time stamps.
+     *
+     * @param start the start time stamp with a time step
+     * @param end the end time stamp
+     * @param step the time step
+     * @return the frames overlapped by the given time window, spaced by a given
+     *         step. The start time doesn't need to be aligned with the step
+     *         and the time interval is open on the right. The result may be
+     *         empty if the time window falls out of the report window.
+     * @throw std::logic_error if invalid interval or stride < timeStep
```
hernando: or step is not multiple of timeStep

tests/brain/python/compartmentReport.py
```diff
@@ -69,6 +92,15 @@ def test_frames(self):
         # falling on a different frame, so we get two frames.
         assert(len(timestamps) == 2)
         assert(frames.shape == (2, 3))
+
+
+        ### load(start,end,stride)
```
hernando: stride->step

brain/compartmentReport.h
```diff
@@ -77,8 +89,13 @@ class CompartmentReport
     BRAIN_API CompartmentReport(const URI& uri);
     BRAIN_API ~CompartmentReport();
 
+#ifndef DOXYGEN_TO_BREATHE
     /** @return the metadata of the report */
     BRAIN_API const CompartmentReportMetaData& getMetaData() const;
+#else
+    /** @return the metadata of the report */
```
hernando: You have forgotten to add the specification of the dictionary returned (key names and semantics)

PR #167, opened 2017-07-19
Fix wheels build with http_proxy casing inconsistencies

tribal-tec:


tribal-tec:
Should build now again on OSX. Please review

PR #169, opened 2017-07-20
Bugfix in test for load step being multiple of report step.

hernando:
Added new tests for exceptions in python wrapping.

### DIFF
--- a/brain/compartmentReportView.cpp
+++ b/brain/compartmentReportView.cpp
@@ -22,6 +22,17 @@
 
 namespace brain
 {
+namespace
+{
+bool _isMultiple(const double a, const double b)
+{
+    const auto remainder = std::fmod(a, b);
+    // Using double epsilon is too strict for large values of a/b
+    const auto epsilon = b * std::numeric_limits<float>::epsilon();
+    return remainder <= epsilon || (b - remainder) <= epsilon;
+}
+}
+
 CompartmentReportView::CompartmentReportView(
     const std::shared_ptr<detail::CompartmentReportReader>& readerImpl,
     const brion::GIDSet& gids)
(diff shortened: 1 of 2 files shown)


brain/compartmentReportView.cpp
```diff
@@ -112,9 +112,13 @@ std::future<brion::Frames> CompartmentReportView::load(double start, double end,
         throw std::logic_error("Invalid step");
 
     // check step is multiple of timestep
-    if (fmod(step, reportTimeStep) > std::numeric_limits<double>::epsilon())
+    const auto remainder = std::fmod(step, reportTimeStep);
```
arsenius7: Would also simplify testing this particular condition.

tests/brain/python/compartmentReport.py
```diff
@@ -75,32 +75,36 @@ def test_mapping(self):
...
-        ### load(start,end,step)
-        timestamps, frames = view.load(0.0,1.0,0.2)
+        # load(start, end, step)
+        timestamps, frames = view.load(0.0, 1.0, 0.2)
         assert(len(timestamps) == 5)
         assert(numpy.isclose(timestamps,[ 0.0, 0.2, 0.4, 0.6, 0.8]).all())
-        assert(frames.shape == (5,3))
+        assert(frames.shape == (5, 3))
+
+        timestamps, frames = view.load(0.0, 1.0, 0.3)
```
hernando: The problem was not the accuracy of the test but its incompleteness. If was working only for cases there step = reportStep * n + epsilon. If step was 0.29999999... (0.3 in double) and reportStep 0.1, the test was failing because fmod was 0.0999999, that is reportStep - epsilon.

arsenius7: Yes, I know, just thought that a checking 1000x would _also_ catch potential problems with accuracy.
Anyway, it was just a suggestion.

hernando: I was doing it

PR #170, opened 2017-07-26
Cleanups

eile:


eile:
Don't review yet

eile:
good to go

brion/morphologyPlugin.h
```diff
@@ -144,6 +141,9 @@ class MorphologyPlugin : public boost::noncopyable
     /** @copydoc brion::Morphology::flush */
     virtual void flush() = 0;
     //@}
+
+protected:
+    MorphologyInitData _data;
```
rdumusc: not sure about this, it seems preferable to keep a pure abstract base class without data members

brion/plugin/compartmentReportBinary.cpp
```diff
@@ -166,10 +166,10 @@ void _initAIOControlBlock(aiocb& block, const AIOReadData& readData)
     block.aio_lio_opcode = LIO_READ;
 }
 
-void _readAsync(aiocb** operations, size_t length)
+void _readAsync(aiocb** operations, const size_t length)
 {
     if (lio_listio(LIO_WAIT, operations, length, nullptr) == -1)
-        throw std::runtime_error("Error in AIO operation setup" +
+        throw std::runtime_error("Error in AIO setup: " +
```
rdumusc: there is a mix between LBTHROW and regular throw in this file (and the next). I am not sure what is the benefit of LBTHROW, but I feels inconsistent.

hernando: LBTHROW outputs to stderr. I use throw when I don't want to pollute the console output, but I it doesn't look like it's the case here.

brion/morphologyPlugin.h
```diff
@@ -40,31 +37,28 @@ class MorphologyInitData : public PluginInitData
...
     {
     }
 
-    MorphologyVersion getVersion() const { return _version; }
-    CellFamily getFamily() const { return _family; }
-protected:
-    const MorphologyVersion _version;
-    const CellFamily _family;
+    MorphologyVersion version;
+    CellFamily family;
```
hernando: Why these changes? There seems to be no reason.

brion/plugin/morphologyHDF5.cpp
```diff
@@ -711,8 +698,7 @@ void MorphologyHDF5::_writeV11Metadata(const MorphologyInitData& initData)
     H5::DataSpace familyDS(1, &dim);
     H5::Attribute familyAttr =
         metadata.createAttribute(_a_family, familyEnum, familyDS);
-    _family = initData.getFamily();
-    familyAttr.write(familyEnum, &_family);
+    familyAttr.write(familyEnum, &_data.family);
```
hernando: This means quite a few conflicts for me...

brion/morphologyPlugin.h
hernando: This is a lot of code changes for not so much gain, true the code is shorter, but I guess you spent more time changing the code than having added the data fields to the upcoming morphology plugin.

I'd prefer if you store just the family and version instead of the full data. Sure, the URI is small, but if it's not needed, why keeping it?

eile: The reason is not saving code but eliminating redundancy (DRY). The version and family where duplicated between the init data and plugin impls, with some plugins (unsurprisingly) not handling the information correctly.

brion/plugin/compartmentReportBinary.cpp
eile: Fixed.

brion/morphologyPlugin.h
eile: Needed to make the class assignable and to update during reading of files.

hernando: Not needed if you just keep the family and version in the morphology plugin class instead of a copy of this object.

brion/plugin/compartmentReportBinary.cpp
rdumusc: hmm, strange naming then :-)

eile:
> I's OK, but I'd still prefer you to not copy the whole init data.

Any specific reason? For provenance reasons it kind of becomes necessary in the zeroeq-based data service branch I'm working on.

hernando:
Yes, call it premature optimization if you want, but the fact that init data contains strings, which require dynamic memory makes me prefer to avoid them. Probably it doesn't matter too much, but these tiny details end up building up in terms of memory fragmentation, cache pollution and overall performance. I've just experienced an unexplained 10% slow down in synapse reading by replacing HDF5++ with HighFive and it's probably related to details such as this one.

If you need it in the ZeroEQ implementation just keep it there.

PR #171, opened 2017-08-03
Ban obsolete morphology repair stage from API

eile:


brion/plugin/morphologyHDF5.cpp
```diff
@@ -204,16 +188,14 @@ Vector4fsPtr MorphologyHDF5::readPoints(MorphologyRepairStage stage) const
         H5::DataSet dataset;
         try
         {
-            if (stage == MORPHOLOGY_UNDEFINED)
-                stage = _stage;
-            dataset = _file.openDataSet("/" + _g_root + "/" + toString(stage) +
-                                        "/" + _d_points);
+            dataset = _file.openDataSet("/" + _g_root + "/" + _stage + "/" +
+                                        _d_points);
```
hernando: You have changed the default semantics here. Are you sure there's no tool using v2 morphologies that would be affected by this change?

tribal-tec: In any case this has to documented/mentioned in the changelog.

PR #172, opened 2017-08-04
zeroeq::Server/Client based morphology loader

eile:


eile:
Retest this please.

eile:
ping?

tests/perf/circuit.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include <brain/brain.h>
+#include <lunchbox/clock.h>
+
+#define BOOST_TEST_MODULE CircuitPerf
+#include <boost/test/unit_test.hpp>
+
+BOOST_AUTO_TEST_CASE(load_morphologies)
+{
+    const brain::Circuit circuit(
+        (brion::URI("/home/eilemann/stable/Brion/BlueConfig")));
```
rdumusc: home folder path? What happens if someone runs ninja Brion-perftests?

tests/perf/circuit.cpp
```diff
@@ -0,0 +1,30 @@
...
+    auto gids = circuit.getGIDs();
+    if (gids.size() > 50000)
```
rdumusc: make this a function, something like:
resizeIfBiggerThan(gids, 50000);

brain/CMakeLists.txt
```diff
@@ -49,14 +49,15 @@ set(BRAIN_SOURCES
...
   PUBLIC Brion vmmlib
-  PRIVATE Keyv ${Boost_FILESYSTEM_LIBRARIES}
+  PRIVATE BrionPlugins Keyv ${Boost_FILESYSTEM_LIBRARIES}
 )
 set(BRAIN_INCLUDE_NAME brain)
 set(BRAIN_NAMESPACE brain)
 
 if(TARGET MVDTool)
-  list(APPEND BRAIN_LINK_LIBRARIES PRIVATE MVDTool)
+  list(APPEND BRAIN_LINK_LIBRARIES MVDTool)
```
eile: done

brion/plugin/morphologyZeroEQ.h
```diff
@@ -0,0 +1,93 @@
...
+
+#include <servus/serializable.h> // base class
+
+#include <future>
+
+namespace brion
+{
+namespace plugin
+{
+class MorphologyZeroEQ : public MorphologyPlugin, public servus::Serializable
```
eile: Let me see what I can do. It's both a serializable morphology (for brain & ZeroEQ) and the ZeroEQ plugin.

brain/neuron/morphologyImpl.cpp
```diff
@@ -76,47 +47,20 @@ Morphology::Impl::Impl(const brion::Morphology& morphology)
     somaSection = ids[0];
 }
 
-bool Morphology::Impl::_fromBinary(const void* data, const size_t size)
+Morphology::Impl::Impl(const brion::Morphology& morphology)
+    : brion::plugin::MorphologyZeroEQ(morphology)
```
eile: ^ see above. Likely I'll move the serializable plugin to a helper class.

tests/perf/circuit.cpp
eile: o_O

brain/circuit.cpp
```diff
@@ -125,24 +128,37 @@ neuron::Morphologies Circuit::loadMorphologies(const GIDSet& gids,
...
+
+        CachedMorphologies::const_iterator it = cached.find(hash);
+        if (it->second)
+            result.push_back(it->second);
+        else
+        {
+            brion::Morphology& raw = loading.find(hash)->second;
+            neuron::MorphologyPtr morphology(
+                transform ? new neuron::Morphology(raw, transforms[i])
+                          : new neuron::Morphology(raw));
```
hernando: You're not updating cached anymore, well, yes, with a nullptr above.

brain/neuron/morphologyImpl.cpp
hernando: This inheritance doesn't neither look right to me at all.
First of all, ZeroEQ is an optional dependency. Given that, making a core class of Brain depend on something that has the ZeroEQ suffix doesn't only smell fishy, is already rotten. Maybe it's just the naming because I haven't reached the class declaration yet, but it could also be that inheritance is used where it should be aggregation.

brion/morphologyPlugin.h
```diff
@@ -35,10 +35,12 @@ namespace brion
...
 public:
-    explicit MorphologyInitData(const URI& uri)
+    explicit MorphologyInitData(
+        const URI& uri, const MorphologyVersion v = MORPHOLOGY_VERSION_H5_1_1,
+        const CellFamily f = FAMILY_NEURON)
         : PluginInitData(uri, MODE_READ)
-        , version(MORPHOLOGY_VERSION_H5_1_1)
-        , family(FAMILY_NEURON)
+        , version(v)
+        , family(f)
```
hernando: This should throw when CellFamiliy is FAMILY_GLIA and the version is not 1.1. The old code was a bit strange exactly to prevent that from being possible at all.

brion/plugin/CMakeLists.txt
```diff
@@ -26,6 +27,7 @@ set(BRIONPLUGINS_SOURCES
   compartmentReportMap.cpp
   morphologyHDF5.cpp
   morphologySWC.cpp
+  morphologyZeroEQ.cpp
```
hernando: The ZeroEQ sources should be added conditionally

tests/morphology.cpp
```diff
@@ -293,6 +297,47 @@ BOOST_AUTO_TEST_CASE(h5_read_v2)
...
+    zeroeq::Server server(zeroeq::NULL_SESSION);
+    server.handle(brion::ZEROEQ_GET_MORPHOLOGY, [](const void* data,
+                                                   const size_t size) {
+        if (!data || !size)
+            return zeroeq::ReplyData();
+
+        const std::string path((const char*)data, size);
+        const brion::Morphology morphology(path);
+        const brion::plugin::MorphologyZeroEQ serializable{morphology};
+        return zeroeq::ReplyData(serializable.getTypeIdentifier(),
```
hernando: I'm confused by this getTypeIdentifier(), it seems that it wouldn't match the code in the server application, where you use ZEROEQ_GET_MORPHOLOGY.

eile:
@rdumusc/all: See last commit for a substantial cleanup in the internals. This breaks the API - shall I bump major?

tests/perf/circuit.cpp
eile: Accidental testing artefact. Fixed.

apps/CMakeLists.txt
```diff
@@ -24,6 +24,13 @@ if(TARGET BBPTestData)
   list(APPEND SPIKECONVERTER_LINK_LIBRARIES BBPTestData)
 endif()
 
+if(TARGET ZeroEQ)
```
eile: common_find_package(Keyv REQUIRED) ?

apps/CMakeLists.txt
```diff
@@ -24,6 +24,13 @@ if(TARGET BBPTestData)
...
+  set(MORPHOLOGYSERVER_HEADERS)
```
eile: done

apps/morphologyServer.cpp
```diff
@@ -0,0 +1,84 @@
...
+    const std::string address = server.getURI().getHost() + ":" +
+                                std::to_string(int(server.getURI().getPort()));
+    std::cout
+        << "Morphology server reachable on:" << std::endl
+        << "  export ZEROEQ_SERVERS=\"" << address << ",$ZEROEQ_SERVERS\""
+        << std::endl
+        << "  zeroeq://" << address << "/path/to/morphology" << std::endl
+        << std::endl
+        << "  [c]ache read, [d]isk read with cache update, disk read "
+           "with cache [e]rror, [D]isk read uncached, morphology [l]oad error: "
```
eile: done

apps/morphologyServer.cpp
```diff
@@ -0,0 +1,84 @@
...
+            cache->takeValues({path},
+                              [&](const std::string&, char* d, const size_t s) {
+                                  value.ptr.reset(d);
+                                  value.size = s;
+                              });
+
+            if (value.ptr && value.size)
+            {
+                std::cout << 'c' << std::flush;
+                return zeroeq::ReplyData(brion::ZEROEQ_GET_MORPHOLOGY, value);
```
eile: It returns the reply ID, not the request ID (which in this use case is identical to the request ID). A handler may return different reply IDs to the same request, e.g., for error codes or different return value types.

brain/circuit.cpp
```diff
@@ -125,24 +128,37 @@ neuron::Morphologies Circuit::loadMorphologies(const GIDSet& gids,
...
         CachedMorphologies::const_iterator it = cached.find(hash);
         if (it == cached.end())
         {
-            neuron::MorphologyPtr morphology;
-            const brion::Morphology raw(uri.getPath());
-            if (transform)
-                morphology.reset(new neuron::Morphology(raw, transforms[i]));
-            else
-                morphology.reset(new neuron::Morphology(raw));
+            // delay data access to next loop to allow async loading
```
eile: done

brain/circuit.cpp
```diff
@@ -125,24 +128,37 @@ neuron::Morphologies Circuit::loadMorphologies(const GIDSet& gids,
...
+        }
+    }
+
+    // load and transform missing and put them in GID-order into result
+    for (size_t i = 0; i < uris.size(); ++i)
+    {
+        const URI& uri = uris[i];
+        const std::string& hash = hashes[i];
+
+        CachedMorphologies::const_iterator it = cached.find(hash);
```
eile: done

apps/CMakeLists.txt
hernando: Where did I looked at then...? I checked it before commenting.

brain/circuit.cpp
eile: good catch. Fixed.

apps/morphologyServer.cpp
hernando: That was not obvious by the code because in case of errors you're returning and empty reply.

brion/morphologyPlugin.h
eile: done.

brion/plugin/CMakeLists.txt
eile: Now yes, after the serialization code is factored out. Done.

tests/morphology.cpp
eile: It doesn't match, but since it's ignored by the reply handler it does not matter. Fixed.

eile:
All comments addressed. Please review new stuff.

brain/neuron/morphologyImpl.cpp
```diff
@@ -31,40 +31,29 @@ namespace brain
...
+Morphology::Impl::Impl(brion::Morphology& morphology)
+    : data(morphology)
+{
+    _extractInformation();
 }
 
 SectionRange Morphology::Impl::getSectionRange(const uint32_t sectionID) const
 {
+    auto points = data.getPoints();
+    auto sections = data.getSections();
```
hernando: You can also add const, which is not automatic if it's not required.

brion/detail/serializableMorphology.ipp
```diff
@@ -0,0 +1,164 @@
...
+                       const uint8_t*& src, const uint8_t* end)
+{
+    if (src + sizeof(uint64_t) > end)
+        return false;
+    const uint64_t arraySize = *reinterpret_cast<const uint64_t*>(src);
+    src += sizeof(uint64_t);
+
+    if (src + sizeof(T) * arraySize > end)
+        return false;
+    const T* dstPtr = reinterpret_cast<const T*>(src);
```
hernando: dstPtr -> strPtr

brion/detail/serializableMorphology.ipp
```diff
@@ -0,0 +1,164 @@
...
+
+void inline SerializableMorphology::setApicals(const Vector2is& apicals)
+{
+    _apicals.reset(new Vector2is(apicals));
+}
+
+void inline SerializableMorphology::setPerimeters(const floats& perimeters)
+{
+    _perimeters.reset(new floats(perimeters));
+}
```
hernando: Does it make sense to add methods that take the vectors as rvalue references and moves them?

brion/detail/serializableMorphology.ipp
```diff
@@ -0,0 +1,164 @@
...
+    ptr += sizeof(MorphologyVersion);
+
+    *reinterpret_cast<CellFamily*>(ptr) = _family;
+    ptr += sizeof(CellFamily);
+
+    _serializeArray(ptr, _points);
+    _serializeArray(ptr, _sections);
+    _serializeArray(ptr, _types);
+    _serializeArray(ptr, _apicals);
+    _serializeArray(ptr, _perimeters);
```
hernando: Consider perimeters only if cell family is glia?

brion/detail/serializableMorphology.ipp
```diff
@@ -0,0 +1,164 @@
...
+        LBTHROW(std::runtime_error(
+            "Failed to construct morphology from binary data"));
+}
+
+inline SerializableMorphology::SerializableMorphology(brion::Morphology& m)
+    : _points(m.readPoints())
+    , _sections(m.readSections())
+    , _types(m.readSectionTypes())
+    , _apicals(m.readApicals())
+    , _perimeters(m.readPerimeters())
```
hernando: Double check that perimeters may not be a nullptr for neurons, this will break your code (but presumably this is already tested)

brion/morphology.cpp
```diff
@@ -69,27 +69,27 @@ CellFamily Morphology::getCellFamily() const
     return _impl->plugin->getCellFamily();
 }
 
-Vector4fsPtr Morphology::readPoints() const
+Vector4fsPtr Morphology::readPoints()
```
hernando: Do we really need to break the API here. The inconsistency in the current code is in the SWC reader, which actually returns private data and makes it shared. The H5 reader is reading the data each time the function is called. If SWC is modified to return a copy there wouldn't be any logical issue with this const. Why it must be different in the ZeroEQ case despite the function returns always the same value (as long as this function doesn't really have visible side effects on the object state it should be considered const).

brion/detail/serializableMorphology.ipp
eile: done

brion/detail/serializableMorphology.ipp
eile: done

brion/detail/serializableMorphology.ipp
eile: No, this detail should not leak out of the plugin. The contract is that some of the arrays may be nullptr and client code should handle this. My last 'done' above did add this here.

brion/morphology.cpp
eile: Both SWC and ZeroEQ then need to create a copy on each call, otherwise it's equivalent to e.g. ```int& getFoo() const``` which is not correct. For example, the brain morphology transforms the points which has a side effect when applied to a SWC brion morphology.

I decided for this since it's more efficient, explicit and sufficient for the current use cases.

brion/detail/serializableMorphology.ipp
hernando: OK, but this is not a plugin detail, but a data model detail. Client code using Brion also needs to take it into account.

brion/detail/serializableMorphology.ipp
eile: Not without braking the brion::Morphology API: The write functions take a const & which is forwarded to the (zeroeq) Plugin using these methods.

hernando: That would break the API but extend it. In any case. This is not a concern so let's keep it as it is.

brion/morphology.cpp
hernando: Then the version number should be bumped for consistency. Even if we are not strict about it at times you should try to be more disciplined about it.

eile: It's already been bumped after the release.

brion/detail/serializableMorphology.ipp
eile: I find
```
   auto perimeters = m.readPerimeters();
   if(perimeters)
     doStuff(perimeters);
```
way more robust, elegant and logical than
```
  if(m.getFamily() == GLIA)
     doStuff(m.readPerimeters());
```

hernando: Depends on the use case, but my point is not about the code, but the statement about it being an implementation detail. Either one way or the other, client code outside Brion also needs to be aware that perimeters may be empty.

eile:
Ping: Please either tell me what is left or approve.

brion/morphology.cpp
hernando: I was referring to the major number.

brion/plugin/morphologyZeroEQ.cpp
```diff
@@ -0,0 +1,229 @@
...
+        ClientPtr client_ = _client; // keep ref for thread-safety
+        while (_client)
+            client_->receive();
+        return _data.hasData();
+    });
+}
+
+MorphologyZeroEQ::~MorphologyZeroEQ()
+{
+    _load();
```
hernando: ???

eile: The future needs to be completed in case the user does not call any read.

brion/plugin/morphologyZeroEQ.cpp
```diff
@@ -0,0 +1,229 @@
...
+    if (!client->request(ZEROEQ_GET_MORPHOLOGY, path.data(), path.size(),
+                         handler))
+    {
+        LBTHROW(std::runtime_error("Failed to request morphology data"));
+    }
+
+    _loader = _getWorkers().post([&] {
+        ClientPtr client_ = _client; // keep ref for thread-safety
+        while (_client)
+            client_->receive();
```
hernando: This is unnecessarily hogging the CPU like hell when loading hundreds of morphologies.  I'd have been failed any exam in concurrent programming had I done this.

As I understand the implementation, receive can handle the request of any morphology, not necessarily this one, which is why you have the while(_client) condition. Having multiple threads making use of the Client doesn't seem very logical on the other hand. I'd guess this is causing a lot of L2 misses and invalidations in the buffers used for network communication. It looks a look more reasonable have a single thread doing the network stuff and having workers for the fromBinary conversions. The worker (which would replace your thread-safe Client) provides a post method that returns a future and the workers wait on that future.

brion/plugin/morphologyZeroEQ.h
```diff
@@ -0,0 +1,66 @@
...
+    void writeApicals(const Vector2is& apicals) final;
+    void writePerimeters(const floats& perimeters) final;
+    void flush() final;
+
+private:
+    // ZeroEQ API
+    class Client;
+    using ClientPtr = std::shared_ptr<Client>;
+    ClientPtr _getClient();
+    void _load(); // throws
```
hernando: throws what and when?

tests/brain/morphology.cpp
```diff
@@ -93,8 +93,8 @@ const std::string TEST_MORPHOLOGY_FILENAME =
 const brion::URI TEST_MORPHOLOGY_URI =
     brion::URI("file://" + TEST_MORPHOLOGY_FILENAME);
 
-void checkEqualMorphologies(const brain::neuron::Morphology& first,
```
hernando: There's no reason why the brain morphology couldn't be const.

brion/plugin/morphologyZeroEQ.cpp
hernando: Then wrap it around a try catch(...) a destructor that throws it's a bad thing: https://akrzemi1.wordpress.com/2011/09/21/destructors-that-throw/

brion/plugin/morphologyZeroEQ.h
```diff
@@ -0,0 +1,66 @@
...
+#include "../detail/serializableMorphology.h" // member
+#include "../morphologyPlugin.h"              // base class
+
+#include <future>
+
+namespace brion
+{
+namespace plugin
+{
+class MorphologyZeroEQ : public MorphologyPlugin
```
rdumusc: It would be nice to have 1-2 line of documentation for the class.

brion/morphology.h
```diff
@@ -37,7 +37,18 @@ class Morphology
...
+    BRION_API Morphology& operator=(Morphology&&);
+
+    /**
+     * @name Read API
+     *
+     * Data returned by the read methods may or may not be cached by the
+     * implementation. If the returned vectors are modified this may be
+     * reflected in subsequent reads of the same data. Furthermore, calling read
+     * multiple times might be expensive. It is therefore recommended to call
+     * each read method at most once per instantiation.
```
eile: Out of the scope for this PR. Imo the motivation is to keep Brion simple by placing as little constraints on the impl as possible. All existing users follow the recommendation anyways (call read once). I tried to got the const route, but this has it's own can of worms which smelled fishy (mutable on unique_ptr, different behaviour if called twice, or memcpy).

brion/plugin/morphologyZeroEQ.cpp
eile: Changed naming.

brion/plugin/morphologyZeroEQ.cpp
eile: Nice horse you've got there. 

Having workers for the fromBinary means you need to memcpy the data out of the zmq_msg which will certainly be slower. We could extend the ZeroEQ API to allow ownership transfer of the zmq messages, but at what cost to resolve one minor itch here? After all, these threads only spin as long as data is loading.

I've done a quick test (e06a167) having only one thread with blocking receive, and thats slower in all cases (17k vs 12k/s locally, 3k vs .8k remotely from Lugano).

BTW: zmq_poll seems to spin too. At least the morphologyServer (with a blocking receive) hogs a CPU too.

brion/plugin/morphologyZeroEQ.h
eile: Done, although in 10 lines.

tribal-tec:
Considering [this](https://bbpcode.epfl.ch/ci/job/oss.Brion.wheel/platform=rh7.x/130/console) and the preview nature of this change, I would suggest to disable the ZeroEQ subproject to keep the wheel building running. Otherwise, the docker image has to integrate boost_thread and libzmq (at least, maybe more).

eile:
@tribal-tec: Agreed, ZeroEQ is removed from .gitsubprojects.

eile:
All: Please review new read API (76bbfa95f2393918a441032f04df59f690c55f9d), removed write API (3a8ff150d2b2c56eb21e5c0323d344e2f6031886) (spec in 802b4e518d4f84642ae55e2d8e0343e31769d653)

.gitsubprojects
```diff
@@ -1,6 +1,5 @@
 # -*- mode: cmake -*-
 git_subproject(Servus https://github.com/HBPVIS/Servus.git 7a60583)
-git_subproject(ZeroEQ https://github.com/HBPVIS/ZeroEQ.git e6f91b0)
```
tribal-tec: Well, just disabling ZeroEQ in docker_build_wheel.sh would have been enough I suppose. Your fix also works of course.

eile:
Ping^2

eile:
> Version bump to major 3?
done

brion/morphology.cpp
```diff
@@ -36,53 +72,143 @@ class Morphology::Impl
...
+    {
+    }
+
+    void finishLoad() const
+    {
+        std::lock_guard<std::mutex> lock(futureMutex);
+        if (!loadFuture.valid())
+            return;
+
+        loadFuture.wait();
```
tribal-tec: rm

PR #181, opened 2017-09-04
Fix in H5 mesh reading.

hernando:


brion/detail/meshHDF5.h
```diff
@@ -138,7 +138,12 @@ class MeshHDF5 : public Mesh
         const auto dataset = _file.getDataSet("/membrane/mesh/triangles/soup");
         uint32_tsPtr buffer(new uint32_ts);
         buffer->resize(_numElements(dataset) * 3);
-        dataset.read(*buffer);
+        // Tricking HighFive into believing this is a 2D array. This is a nasty
+        // hack, but it works because HighFive is casting the pointer back
+        // to void* before copying the data read into it. The truth is that
+        // using a double * to mean 2 dimensions seems utterly broken
```
eile: oh, double * as in 2x * not pointer-to-double. Maybe just write as using a '**' to ...?

hernando:
I'm fixing the behaviour in HighFive

PR #186, opened 2017-10-11
Do not post from a thredPool thread into the same pool

hernando:
This was done in the compartment report reading code and can lead
to a trivial deadlock if the pool doesn't have enough threads that
can finish their tasks before the new task is posted.
A trivial case is when the hardware concurrency is 1.


hernando:
ping

PR #187, opened 2017-10-16
Clang9

hernando:


hernando:
Latest Travis osx default version is already 9.3

PR #188, opened 2017-10-18
Allow Keyv to be optional

hernando:


brain/CMakeLists.txt
```diff
@@ -49,8 +49,13 @@ set(BRAIN_SOURCES
 set(BRAIN_PUBLIC_INCLUDE_DIRECTORIES ${Boost_INCLUDE_DIRS})
 set(BRAIN_LINK_LIBRARIES
   PUBLIC Brion vmmlib
-  PRIVATE Keyv ${Boost_FILESYSTEM_LIBRARIES}
+  PRIVATE Lunchbox ${Boost_FILESYSTEM_LIBRARIES}
 )
+
+if(TARGET Keyv)
+  list(APPEND BRAIN_LINK_LIBRARIES Keyv)
```
rdumusc: I find it better to repeat the PRIVATE keyword here than rely on context

brion/plugin/CMakeLists.txt
```diff
@@ -33,10 +31,16 @@ set(BRIONPLUGINS_SOURCES
...
 set(BRIONPLUGINS_LINK_LIBRARIES
-  PRIVATE Lunchbox Keyv Servus vmmlib HighFive ${Boost_FILESYSTEM_LIBRARIES}
+  PRIVATE Lunchbox Servus vmmlib HighFive ${Boost_FILESYSTEM_LIBRARIES}
           ${Boost_REGEX_LIBRARIES} ${CMAKE_THREADS_LIB_INIT}
 )
 
+if(TARGET Keyv)
+  list(APPEND BRIONPLUGINS_HEADERS compartmentReportMap.h)
+  list(APPEND BRIONPLUGINS_SOURCES compartmentReportMap.cpp)
+  list(APPEND BRIONPLUGINS_LINK_LIBRARIES Keyv)
```
rdumusc: same

brain/CMakeLists.txt
hernando: I did at first, but I found that it was not done ZeroEQ. I'll do it for both

PR #190, opened 2017-10-23
Update ZeroEQ to be able to build Brayns with Livre + Brion

rdumusc:


hernando:
Looks like you need to update Servus

PR #191, opened 2017-11-14
Report optimizations

hernando:


brion/compartmentReportPlugin.h
```diff
@@ -45,16 +45,34 @@ class CompartmentReportInitData : public PluginInitData
      * @param gids the neurons of interest in READ_MODE
      * @version 1.4
      */
-    explicit CompartmentReportInitData(const URI& uri,
-                                       const int accessMode = MODE_READ,
+    explicit CompartmentReportInitData(const URI& uri, const int accessMode,
```
tribal-tec: no explicit; technically also now version 3.0

brion/plugin/compartmentReportBinary.cpp
```diff
@@ -317,8 +331,18 @@ bool CompartmentReportBinary::_remapFile(const size_t size)
     return _file.is_open();
 }
 
+size_t CompartmentReportBinary::getCellCount() const
+{
+    if (_gids.empty())
+        return _header.numCells;
+    else
```
tribal-tec: rm else

brion/plugin/compartmentReportBinary.cpp
```diff
@@ -662,29 +689,31 @@ bool CompartmentReportBinary::_parseHeader()
     if (_tunit.empty())
         _tunit = "ms";
 
+    // After parsing the header we figure out the end of the mapping by reading
+    // the data offset of the first cell.
+    _dataOffset = get<uint64_t>(ptr, DATA_INFO + _header.headerSize);
```
tribal-tec: why did this move out of _parseMapping()? It's always the same for multiple updateMapping() calls

brion/plugin/compartmentReportBinary.cpp
```diff
@@ -662,29 +689,31 @@ bool CompartmentReportBinary::_parseHeader()
...
+        auto gid = get<int32_t>(ptr, i * SIZE_CELL_INFO_LENGTH);
+        lunchbox::byteswap(gid);
+        _originalGIDs.insert(gid);
     }
+}
 
+bool CompartmentReportBinary::_parseMapping()
+{
+    const uint8_t* ptr = reinterpret_cast<const uint8_t*>(_file.data());
+    size_t offset = _header.headerSize;
```
tribal-tec: const?

brion/compartmentReportPlugin.h
hernando: ok

brion/plugin/compartmentReportBinary.cpp
hernando: done

brion/plugin/compartmentReportBinary.cpp
hernando: done

hernando:
I also removed cellCount from the metadata.
I'm actually wondering if it makes sense to keep this metadata struct (C++) / dictionary (python) or should we make all metadata functions/properties of the CompartmentReport class (@arsenius7, @tribal-tec  opinions?)

hernando:
I made the change because even H5Gget_num_objs is too slow.

arsenius7:
> I'm actually wondering if it makes sense to keep this metadata struct (C++) / dictionary (python) or should we make all metadata functions/properties of the CompartmentReport class

As we discussed yesterday, I would propose to distinguish between 'required' and 'optional' metadata.
For 'required' fields (for BluePy these are now start_time, end_time, time_step), it makes perfect sense to move them to standalone methods.
For the optional ones I would keep them in the dict and relax any requirements on it (fixed key set, fast response etc). They could still be useful for debugging so I would suggest to keep them.

hernando:
In fact, if I knew how to translate dict lookups in Python into function calls in C++ I wouldn't change the Python side. For the C++ side is not the same, as it doesn't provide the required language feature.

brion/plugin/compartmentReportBinary.cpp
hernando: (I thought I had already answered this, but I can't see the comment)
Yes, in fact it should the data offset should be computable just with the information from the header, but it doesn't happen to be the case (or I didn't spend enough time figuring it out).

hernando:
ping

brion/plugin/compartmentReportBinary.cpp
tribal-tec: still no satisfying answer why you had to move the code. Where it was before should still work, no?

hernando: It can't be at the original place. The offset is needed for remapping the file (line 282) and then read the GIDs, which is done even if the full mapping is not processed.

PR #198, opened 2017-12-04
Do no link Boost.iostreams against bzip2 in the Python wheel

hernando:


hernando:
I'm afraid that unless they have fixed the hdf5 packages in homebrew, the Travis builds are going to fail.

PR #200, opened 2017-12-06
Fixed a regression introduced in e84ff45

hernando:
Parsing of gids from little endian binary reports was wrong.


brion/plugin/compartmentReportBinary.cpp
```diff
@@ -705,7 +705,8 @@ void CompartmentReportBinary::_parseGIDs()
     for (int32_t i = 0; i < _header.numCells; ++i)
     {
         auto gid = get<int32_t>(ptr, i * SIZE_CELL_INFO_LENGTH);
-        lunchbox::byteswap(gid);
+        if (_header.byteswap)
```
tribal-tec: ouch!

hernando: yes... and that's why unit tests are not the holy grail. If we had to test all the possible combinations they would be trillions. On the other hand, this should have been spotted during the review.

PR #204, opened 2018-01-17
Fixed brain::circuit methods to not crash with empty gids lists

hernando:


tests/brain/circuit.cpp
```diff
@@ -188,6 +188,24 @@ BOOST_AUTO_TEST_CASE(brain_circuit_positions)
         0.000001f);
 }
 
+BOOST_AUTO_TEST_CASE(brain_empty_gids_lists)
+{
+    const brain::Circuit mvd2((brion::URI(bbp::test::getBlueconfig())));
+    brion::GIDSet nil;
+    BOOST_CHECK(mvd2.getPositions(nil).empty());
+    BOOST_CHECK(mvd2.getTransforms(nil).empty());
+    BOOST_CHECK(mvd2.getMorphologyNames(nil).empty());
```
tribal-tec: getRotations() is missing?

hernando: getTransforms uses the the same code path than getRotations, but I can add it.

PR #212, opened 2018-03-13
Sonata data parsing

karjonas:
This is a work in progress for parsing the sonata files. It contains the parsing of the H5-file with test cases on a sample file.


brion/nodeGroup.h
```diff
@@ -0,0 +1,64 @@
...
+
+namespace brion
+{
+class Node;
+
+class NodeGroup
+{
+public:
+    NodeGroup(const NodeGroup& other);
+    NodeGroup(const HighFive::Group& group);
```
hernando: Should be private and Nodes friend of this class

brion/nodeGroup.h
```diff
@@ -0,0 +1,64 @@
...
+}
+
+namespace brion
+{
+class Node;
+
+class NodeGroup
+{
+public:
+    NodeGroup(const NodeGroup& other);
```
hernando: Needed?

brion/nodeGroup.h
```diff
@@ -0,0 +1,64 @@
...
+    NodeGroup(const HighFive::Group& group);
+    ~NodeGroup();
+
+    Strings getAttributeNames();
```
hernando: Documentation missing. You could, for example, state that this function traverses all the datasets in the node group to find out the names. The user should have an understanding of the cost of the function apart from what it returns.

brion/nodeGroup.h
```diff
@@ -0,0 +1,64 @@
...
+
+    Strings getDynamicParameterNames();
```
hernando: Same as above

brion/nodeGroup.h
```diff
@@ -0,0 +1,64 @@
...
+
+    size_t getNumberOfNodes() const;
+
+    template <typename T>
+    T getAttribute(const std::string& name) const;
```
hernando: We made a mistake here. The return value should be std::vector<T>. The same applies to all the functions below.

Document what happens when the name is wrong or the conversion is not possible.

brion/nodes.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+#include <HighFive/include/highfive/H5DataSet.hpp>
+#include <HighFive/include/highfive/H5File.hpp>
+
+#include <iostream>
+#include <memory>
+
+namespace brion
+{
+static uint32_ts get_id_list_helper(const HighFive::File& file,
```
hernando: Use an anonymous namespace instead of static.
The coding convention would be _getIDList (no helper needed, it's obvious it's a helper functions).

The name is a bit deceiving, this is not extracting necessarily an id list but reading a dataset into a uin32_t vector, I would call it _readIntVector

brion/nodes.cpp
```diff
@@ -0,0 +1,114 @@
...
+{
+    std::unique_ptr<HighFive::File> file;
+};
+
+Nodes::Nodes(const URI& uri)
+    : impl(new Nodes::Impl())
+{
+    const std::string path = uri.getPath();
+
+    try
```
hernando: Do you think you really need to catch the exception to rethrow?
It's less lines of code if you do something like:

    : impl(new Node::Impl(uri.getPath())

And then you can also save the pointer in Nodes::Impl

It's also not consistent that you are catching the HighFive exception here, but not in any other function.

brion/nodes.cpp
```diff
@@ -0,0 +1,114 @@
...
+        throw std::runtime_error("Could not open file " + path + ": " +
+                                 exc.what());
+    }
+}
+
+Nodes::~Nodes() = default;
+
+Strings Nodes::getPopulationNames() const
+{
+    Strings population_names;
```
hernando: Not a big deal, but we use camelCase for all names.

brion/nodes.cpp
```diff
@@ -0,0 +1,114 @@
...
+    return get_id_list_helper(*impl->file.get(), population, "node_group_id");
+}
+
+uint32_ts Nodes::getNodeGroupIndices(const std::string population) const
+{
+    return get_id_list_helper(*impl->file.get(), population,
+                              "node_group_index");
+}
+
+uint32_ts Nodes::getNodeTypes(const std::string population) const
```
hernando: According to the discussion we had with ABI, this is going to stay at the top level group.

brion/nodes.cpp
```diff
@@ -0,0 +1,114 @@
...
+{
+    return impl->file->getGroup("/nodes/" + population)
+        .getDataSet("node_type_id")
+        .getSpace()
+        .getDimensions()[0];
+}
+
+uint32_ts Nodes::getNodeIDs(const std::string population) const
+{
+    return get_id_list_helper(*impl->file.get(), population, "node_id");
```
hernando: I think you can simply type *impl->file

brion/nodes.h
```diff
@@ -0,0 +1,69 @@
...
+        population.
+        @throw runtime_error if the dataset does not exist. */
+    uint32_ts getNodeGroupIDs(const std::string population) const;
+
+    /** Read and return the values in the node_group_ind dataset of the
+        requested population.
+        @throw runtime_error if the dataset does not exist. */
+    uint32_ts getNodeGroupIndices(const std::string population) const;
+
+    /** TBD: may move to the group level. */
```
hernando: remove this comment and write proper documentation.

brion/nodes.h
```diff
@@ -0,0 +1,69 @@
...
+
+#include <memory>
+
+namespace brion
+{
+class Nodes
+{
+public:
+    /** Open a nodes h5 file.
+        @throw runtime_error if the file couldn't be opened or it's not a node
```
hernando: According to the comment I made above, this would need to change.

brion/nodeGroup.cpp
```diff
@@ -0,0 +1,185 @@
...
+                                 size_t end) const
+{
+    return std::move(getAttributeHelper<T>(
+        impl->group.getGroup("dynamics_params").getDataSet(name), true, start,
+        end));
+}
+
+//////////////////////////////////////////////////////////////////////////////
+
+template int32_ts NodeGroup::getAttribute<int32_ts>(const std::string&) const;
```
hernando: I would use a #define to instantiate all the templates for a given type, it wil look shorter.

brion/nodeGroup.cpp
```diff
@@ -0,0 +1,185 @@
...
+}
+
+size_t NodeGroup::getNumberOfNodes() const
+{
+    return impl->group.getNumberObjects();
+}
+
+template <typename T>
+T getAttributeHelper(const HighFive::DataSet set, bool use_slice = false,
+                     size_t start = 0, size_t end = 0)
```
hernando: You don't need the bool parameter if end = std::numeric_limits<size_t>::max()

brion/nodeGroup.cpp
```diff
@@ -0,0 +1,185 @@
...
+{
+    T list;
+    set.read(list);
```
hernando: This is not efficient for the slice case, you're reading everything. You have to use the function HighFive::SliceTraits::select, which is inherited by HighFive::DataSet

brion/nodeGroup.cpp
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    return list;
+}
+
+template <typename T>
+T NodeGroup::getAttribute(const std::string& name, size_t start,
+                          size_t end) const
+{
+    return std::move(
```
hernando: As far as I know, this std::move is unnecessary. The return value of the helper function is already an rvalue in this context.

brion/nodeGroup.h
```diff
@@ -0,0 +1,64 @@
...
+
+    template <typename T>
+    T getAttribute(const std::string& name, size_t start, size_t end) const;
```
hernando: Document:
* What happens when start > end?
* Is the interval open or closed on the right?
* What happens when end is beyond the end of the dataset.

tests/sonata.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+#define BOOST_TEST_MODULE Sonata
+#include <boost/test/unit_test.hpp>
+#include <cstdarg>
+
+// typedef for brevity
+typedef brion::Vector4f V4f;
+typedef brion::Vector3f V3f;
+
+const brion::URI TEST_MORPHOLOGY_URI(std::string("file://") + BRION_TESTDATA +
```
hernando: C&P typo

tests/sonata.cpp
```diff
@@ -0,0 +1,181 @@
...
+static bool fuzz_eq(float v0, float v1)
+{
+    constexpr auto epsilon = 0.00001;
+    return std::abs(v0 - v1) < epsilon;
+}
+
+BOOST_AUTO_TEST_CASE(sonata_constructors)
+{
+    brion::Nodes nodes0(TEST_MORPHOLOGY_URI);
+    brion::Nodes nodes1(TEST_MORPHOLOGY_URI);
```
hernando: What does it add the second call?

hernando: Test openning an unexistent file

tests/sonata.cpp
```diff
@@ -0,0 +1,181 @@
...
+    const auto rotation_angle_y =
+        group.getAttribute<brion::floats>("rotation_angle_y");
+    const auto rotation_angle_z =
+        group.getAttribute<brion::floats>("rotation_angle_z");
+    const auto x = group.getAttribute<brion::floats>("x", 2, 12);
+    const auto y = group.getAttribute<brion::floats>("y", 3, 12);
+    const auto z = group.getAttribute<brion::floats>("z", 6, 10);
+
+    BOOST_CHECK_EQUAL(rotation_angle_x.size(), 10);
+    BOOST_CHECK(fuzz_eq(rotation_angle_x[0], 0.1));
```
hernando: Use BOOST_CHECK_EQUAL and use 0.1f. There are not arithmetic operations or conversions involved, so the values should be bitwise equal.
And by the way, there's a BOOST_CHECK_CLOSE.

PR #218, opened 2018-03-20
Fixing Linux Python wheel build

hernando:


hernando:
This doesn't seem to be sufficient, there's something strange going, probably caused by the paths that have changed.

PR #219, opened 2018-03-22
Sonata Csv file parsing

karjonas:


tests/sonata.cpp
```diff
@@ -216,3 +220,19 @@ BOOST_AUTO_TEST_CASE(circuit_config_getNetworkEdges)
...
 }
+
+////////////////////////////////////////////////////////////////////////////
+
+BOOST_AUTO_TEST_CASE(csv_config_constructors)
+{
+    auto csv = brion::CsvConfig(TEST_SONATA_NODE_TYPES_URI);
+}
+
+BOOST_AUTO_TEST_CASE(csv_config_get_property)
```
hernando: Test non existent node_type_ids and names

brion/csvConfig.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <regex>
+#include <streambuf>
+#include <string>
+#include <unordered_map>
+
+namespace
+{
+// Taken from:
+// https://stackoverflow.com/questions/890164/how-can-i-split-a-string-by-a-delimiter-into-an-array
+std::vector<std::string> explode(const std::string& str, const char& ch)
```
hernando: Use this, https://theboostcpplibraries.com/boost.tokenizer I think you can support multiple spaces this way.

Another simple approach is to use a std::stringstream and use << to extract strings as it skips white space by default.

brion/csvConfig.cpp
```diff
@@ -0,0 +1,145 @@
...
+        file.seekg(0, std::ios::end);
+        contents.reserve(file.tellg());
+        file.seekg(0, std::ios::beg);
+
+        contents.assign((std::istreambuf_iterator<char>(file)),
+                        std::istreambuf_iterator<char>());
+
+        const auto lines = explode(contents, '\n');
+        for (const auto& line : lines)
+            table.push_back(explode(line, ' '));
```
hernando: You can simplify all the code above by using the free function std::getline:
http://en.cppreference.com/w/cpp/string/basic_string/getline

brion/csvConfig.cpp
```diff
@@ -0,0 +1,145 @@
...
+            size_t ctr = 0;
+            for (const auto& columnName : table.front())
+            {
+                nameToColumnIndex[columnName] = ctr;
+                ctr++;
+            }
+
+            for (size_t i = 1; i < table.size(); i++)
+            {
+                const auto& nodeTypeIdStr = table[i][0];
```
hernando: You're making the assumption that the nodeTypeId column is the first one. For safety I wouldn't assume it and use nameToColumnIndex instead.

brion/csvConfig.cpp
```diff
@@ -0,0 +1,145 @@
...
+            }
+        }
+
+        { // Fill access maps
+            size_t ctr = 0;
+            for (const auto& columnName : table.front())
+            {
+                nameToColumnIndex[columnName] = ctr;
+                ctr++;
+            }
```
hernando: Remove the node_type_id column after you figure out it's position?

brion/csvConfig.cpp
```diff
@@ -29,36 +30,14 @@
...
-        else
-        {
-            // Accumulate the next character into the sequence
-            next += *it;
-        }
-    }
-    if (!next.empty())
-        result.push_back(next);
-    return result;
+    const boost::char_separator<char> sep{" "};
```
hernando: Could you add additional whitespace in the test file to test that the tokenizer drops empty tokens between whitespaces.

brion/csvConfig.cpp
```diff
@@ -73,43 +52,43 @@ struct CsvConfig::Impl
...
-
-        { // Verify layout
-            const size_t numColumns = table.front().size();
-            for (const auto& row : table)
-            {
-                if (numColumns != row.size())
-                    throw std::runtime_error(
-                        "Number of columns inconsistent in `" + uri + "`");
-            }
+            table.push_back(explode_whitespace(line));
```
hernando: You can merge the two loops and get rid of the line vector.

brion/csvConfig.cpp
```diff
@@ -73,43 +52,43 @@ struct CsvConfig::Impl
...
+            nameToColumnIndex[columnName] = ctr;
+            ctr++;
+        }
+
+        const auto nodeTypeIdIt = nameToColumnIndex.find("node_type_id");
+
+        if (nodeTypeIdIt == nameToColumnIndex.end())
+            throw std::runtime_error("Could not find column 'node_type_id'");
+
+        const auto pos = std::distance(nameToColumnIndex.begin(), nodeTypeIdIt);
```
hernando: This is not only more complicated that needed, but also incorrect. I'm even surprised the tests are passing.

    const auto pos = *nodeTypeIdIt

brion/csvConfig.cpp
```diff
@@ -124,6 +103,26 @@ CsvConfig::CsvConfig(const URI& uri)
 {
 }
 
+size_ts CsvConfig::getNodeTypeIds() const
+{
+    size_ts output;
+    for (const auto& it : impl->nodeTypeIdToRowIndex)
+        output.push_back(it.first);
+    std::sort(output.begin(), output.end());
```
hernando: If they are sorted, document it in the API, otherwise remove this and say they are not guaranteed to be in any order.

brion/csvConfig.cpp
```diff
@@ -124,6 +103,26 @@ CsvConfig::CsvConfig(const URI& uri)
...
+}
+
+Strings CsvConfig::getProperties() const
+{
+    Strings output;
+    for (const auto& it : impl->nameToColumnIndex)
+        output.push_back(it.first);
+    output.erase(std::remove(output.begin(), output.end(), "node_type_id"),
+                 output.end());
+    std::sort(output.begin(), output.end());
```
hernando: Same comment as above.

brion/csvConfig.cpp
karjonas: Ah yes, I am blind :)

brion/csvConfig.cpp
karjonas: Yeah, a miracle this worked in the tests...

brion/csvConfig.h
```diff
@@ -42,10 +42,10 @@ class CsvConfig
     CsvConfig(const CsvConfig& other) = delete;
     ~CsvConfig();
 
-    /** Returns all available node_type_id's */
+    /** Returns all available node_type_id's in random order */
```
hernando: It's not random because it doesn't change everytime you call it, it's just undefined

hernando:
retest this please

PR #222, opened 2018-04-06
Add incomplete Sonata Circuit

karjonas:
It only has support for getPositions and getRotations so far.


tests/data/sonata/simple_nodes.py
```diff
@@ -89,7 +93,7 @@ class ModelType:
     "etype": ["fast", "slow", "fast"]}
 
 writer = csv.writer(out, delimiter=' ',
-                    quotechar='|', quoting=csv.QUOTE_MINIMAL)
+                    quotechar='|', quoting=csv.QUOTE_MINIMAL, lineterminator='\n')
```
hernando: Why did you have to add lineterminator

brain/detail/circuit.h
```diff
@@ -220,14 +187,313 @@ class Circuit::Impl
...
+                expectedIdx++;
+            }
+        }
+    }
+    virtual ~SonataCircuit() {}
+    virtual size_t getNumNeurons() const
+    {
+        LBUNIMPLEMENTED;
+        return 0;
+    }
```
hernando: You can implement this function also

brain/detail/circuit.h
```diff
@@ -220,14 +187,313 @@ class Circuit::Impl
...
+            output.push_back(rotations[gid]);
+        return output;
+    }
+    virtual Strings getMorphologyNames(const GIDSet& /*gids*/) const
+    {
+        LBUNIMPLEMENTED;
+        return Strings();
+    }
+
+    virtual URI getMorphologyURI(const std::string& /*name*/) const
```
hernando: For this function I've just realized that you can almost share the same implementation for all subclasses. You need to add a new virtual function getMorphologySource to replace the _morphologySource member variable in the original implementation.

brain/detail/circuit.h
```diff
@@ -220,14 +187,313 @@ class Circuit::Impl
...
+        return Strings();
+    }
+    virtual Quaternionfs getRotations(const GIDSet& gids) const
+    {
+        Quaternionfs output;
+        for (auto gid : gids)
+            output.push_back(rotations[gid]);
+        return output;
+    }
+    virtual Strings getMorphologyNames(const GIDSet& /*gids*/) const
```
hernando: This needs to be implemented

brain/detail/circuit.h
```diff
@@ -220,14 +187,313 @@ class Circuit::Impl
...
+    virtual const URI& getSynapseSource() const = 0;
+};
+
+class SonataCircuit : public Circuit::Impl
+{
+public:
+    explicit SonataCircuit(const URI& uri)
+        : config(uri)
+        , basePath(boost::filesystem::path(uri.getPath()).parent_path())
+    {
```
hernando: The idea is to open the node group in the constructor, but you shouldn't read any datasets here. The functions that read the datasets must read at most, the slice of data comprised between the lowest and largest gids. Look into MVD3::getPositions to get an idea.

tests/sonata.cpp
```diff
@@ -262,3 +264,93 @@ BOOST_AUTO_TEST_CASE(csv_config_getProperties)
     BOOST_CHECK_EQUAL(properties[2], "mtype");
     BOOST_CHECK_EQUAL(properties[3], "population");
 }
+/////////////////////////////////////////////////////////////////////////////
+
+BOOST_AUTO_TEST_CASE(sonata_SonataConfig_constructors)
+{
+    brain::Circuit circuit(TEST_SONATA_SIMPLE_NETWORK_URI);
```
hernando: Check an exception is thrown is opening a json file which doesn't exist.

brain/detail/circuit.h
```diff
@@ -222,12 +189,335 @@ class Circuit::Impl
...
+        const auto& subnetworks = config.getNodes();
+        const auto& node = subnetworks.front();
+        brion::Nodes nodes(URI(basePath.string() + "/" + node.elements));
```
hernando: This is wrong, you're assuming that paths are relative, which may not be the case.
The CircuitConfig should deal with all path substitutions and return absolute paths so client code doesn't need to deal with this stuff. As a matter of fact, I missed that you haven't dealt with relative paths at all in the CircuitConfig implementation.

brain/detail/circuit.h
```diff
@@ -222,12 +189,335 @@ class Circuit::Impl
...
+
+        if (subnetworks.size() > 1)
+        {
+            LBWARN << "More than one subnetwork found, ignored." << std::endl;
```
hernando: More than one subnetwork found, ignoring extra ones.

brain/detail/circuit.h
```diff
@@ -222,12 +189,335 @@ class Circuit::Impl
...
+        }
+
+        const auto populations = nodes.getPopulationNames();
+        if (populations.size() > 1)
+        {
+            LBWARN << "More than one population found, ignored." << std::endl;
```
hernando: Same as above

brain/detail/circuit.h
```diff
@@ -222,12 +189,335 @@ class Circuit::Impl
...
+        const auto population = populations.front();
+        const auto nodeGroupIDs = nodes.getNodeGroupIDs(population);
+        const auto nodeIDs = nodes.getNodeIDs(population);
+        const auto nodeGroupIndices = nodes.getNodeGroupIndices(population);
+        const size_t numNodes = nodeIDs.size();
+
+        if (nodeGroupIDs.size() > 1)
+        {
+            LBWARN << "More than one group ID found, ignored." << std::endl;
+        }
```
hernando: This dataset has as many elements as cells. You're always going to give a warning unless your circuit has 1 neuron.
What you must check is that all IDs are 0 and throw if there's any other value.

brain/detail/circuit.h
```diff
@@ -222,12 +189,335 @@ class Circuit::Impl
...
+
+        nodeGroup = nodes.openGroup(population, 0);
```
hernando: Move after the the consistency check.

brain/detail/circuit.h
```diff
@@ -222,12 +189,335 @@ class Circuit::Impl
...
+        LBUNIMPLEMENTED;
+        return *(brion::Synapse*)(delete_this);
+    }
+    virtual const URI& getSynapseSource() const
+    {
+        LBUNIMPLEMENTED;
+        return *(URI*)(delete_this);
+    }
+
+    int* delete_this;
```
hernando: Yes, delete it.

brain/detail/circuit.h
```diff
@@ -222,12 +189,335 @@ class Circuit::Impl
...
+    }
+    virtual void _findSynapsePositionsColumns() const
+    {
+        LBUNIMPLEMENTED;
+        return;
+    }
+    virtual const brion::SynapseSummary& getSynapseSummary() const
+    {
+        LBUNIMPLEMENTED;
+        return *(brion::SynapseSummary*)(delete_this);
```
hernando: Use a static variable for all these return values.

brain/detail/circuit.h
```diff
@@ -222,12 +189,335 @@ class Circuit::Impl
...
+                                                endIdx);
+        Strings output;
+        for (const auto gid : gids)
+            output.push_back(names[gid - startIdx]);
+
+        return output;
+    }
+
+    virtual URI getMorphologySource() const final { return morphologySource; }
+    virtual const brion::URI& getCircuitSource() const
```
hernando: The code for morphology caching needs to be reviewed. In principle I think this function should be renamed and return the circuit config .json path, but I have to think about it more carefully.

brain/detail/circuit.h
```diff
@@ -222,12 +189,335 @@ class Circuit::Impl
...
+    {
+        if (gids.empty())
+            return Strings();
+
+        const size_t startIdx = *gids.begin();
+        const size_t endIdx = *gids.rbegin() + 1;
+
+        const auto names =
+            nodeGroup.getAttribute<std::string>("morphology_file", startIdx,
+                                                endIdx);
```
hernando: If this attribute doesn't exist, you have to look up in the CSV file based on the node type. This is exactly the case of ABI's circuits.

brain/detail/circuit.h
```diff
@@ -222,12 +189,335 @@ class Circuit::Impl
...
+        return output;
+    }
+    virtual size_ts getMTypes(const GIDSet& /*gids*/) const
+    {
+        LBUNIMPLEMENTED;
+        return size_ts();
+    }
+    virtual Strings getMorphologyNames() const
+    {
+        return nodeGroup.getAttribute<std::string>("morphology_file");
```
hernando: UNIMPLEMENTED, the name is a bit misleading, it's not the morphology file names, it's the morphology type names. Which we don't have in ABI's circuits as far as I know.

brain/detail/circuit.h
```diff
@@ -222,12 +189,335 @@ class Circuit::Impl
...
+            if (expectedIdx != nodeGroupIndex)
+            {
+                LBTHROW(
+                    std::runtime_error("Expected node group index '" +
+                                       std::to_string(expectedIdx) + "' got '" +
+                                       std::to_string(nodeGroupIndex) + "'"));
+            }
+
+            expectedIdx++;
+        }
```
hernando: I would just compare nodeGroupIndices to a vector created with std::iota, it's shorter and it doesn't really matter which index is out of order because later we have to implement the indirect code.

brain/detail/circuit.h
```diff
@@ -222,12 +189,335 @@ class Circuit::Impl
 
     URI getMorphologyURI(const std::string& name) const
     {
-        URI uri(_morphologySource);
+        URI uri(getMorphologySource());
```
hernando: This is incorrect, but leave it like that I'll fix it later.

brain/detail/circuit.h
karjonas: I cannot use a variable for it since Synapse constructor will throw an exception if I try to open a nonexistent file.

karjonas: Yes, that is what I will do.

brain/detail/circuit.h
karjonas: I do not understand how to do that so I will leave like this since we are removing it anyway

brain/detail/circuit.h
```diff
@@ -214,20 +183,384 @@ class Circuit::Impl
 
     virtual Vector3fs getPositions(const GIDSet& gids) const = 0;
     virtual size_ts getMTypes(const GIDSet& gids) const = 0;
-    virtual Strings getMorphologyNames() const = 0;
+    virtual Strings getMorphologyTypeNames() const = 0;
     virtual size_ts getETypes(const GIDSet& gids) const = 0;
     virtual Strings getElectrophysiologyNames() const = 0;
```
hernando: Rename this one also for consistency

brain/detail/circuit.h
```diff
@@ -214,20 +183,384 @@ class Circuit::Impl
...
 
     URI getMorphologyURI(const std::string& name) const
     {
-        URI uri(_morphologySource);
+        if (boost::filesystem::path(name).is_absolute())
+            return URI(name);
+
+        URI uri(getMorphologySource());
         uri.setPath(uri.getPath() + "/" + name + ".h5");
         return uri;
```
hernando: I'll review this and change it myself. I will commit on top of your branch head.

brain/detail/circuit.h
```diff
@@ -214,20 +183,384 @@ class Circuit::Impl
...
+    virtual void _findSynapsePositionsColumns() const = 0;
+    virtual const brion::SynapseSummary& getSynapseSummary() const = 0;
+    virtual const brion::Synapse& getSynapseAttributes(
+        const bool afferent) const = 0;
+    virtual const brion::Synapse& getAfferentProjectionAttributes(
+        const std::string& name) const = 0;
+    virtual const brion::Synapse* getSynapseExtra() const = 0;
+    virtual const brion::Synapse& getSynapsePositions(
+        const bool afferent) const = 0;
+    virtual const URI& getSynapseSource() const = 0;
```
hernando: I think I'm going to do another some other refactorings here also. This design is looking terrible to me. This will also remove the need for the compilation hacks you made below.

brain/detail/circuit.h
```diff
@@ -214,20 +183,384 @@ class Circuit::Impl
...
+
+        const size_t startIdx = *gids.begin();
+        const size_t endIdx = *gids.rbegin() + 1;
+
+        const std::string propertyName = "morphology_file";
+
+        const auto attributeNames = nodeGroup.getAttributeNames();
+        const bool hasMorpholgyFile =
+            std::find(attributeNames.begin(), attributeNames.end(),
+                      propertyName) != attributeNames.end();
```
hernando: What about adding a function to search for an attribute name in NodeGroup?

hernando:
retest this please

hernando:
Retest this please

PR #223, opened 2018-04-11
Fix ambiguous call error

karjonas:
error: call of overloaded ‘DataSpace(<brace-enclosed initializer list>)’ is ambiguous


brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -478,7 +478,8 @@ HighFive::DataSet CompartmentReportHDF5::_createDataset(const uint32_t gid,
         reportGroup.createDataSet<float>(mappingDatasetName,
                                          HighFive::DataSpace({1, compCount}));
     HighFive::DataSet dataDataset = reportGroup.createDataSet<float>(
-        dataDatasetName, HighFive::DataSpace({numSteps, compCount}));
+        dataDatasetName,
+        HighFive::DataSpace(std::vector<size_t>{numSteps, compCount}));
```
hernando: I had to do this fix for other compilers as well.

PR #226, opened 2018-04-25
Allow to open BlueConfig from memory buffer

tribal-tec:


brion/blueConfig.cpp
```diff
@@ -87,12 +87,17 @@ class BlueConfig
...
+        if (boost::filesystem::is_regular_file(source))
+        {
+            std::ifstream file(source.c_str());
+            if (!file.is_open())
+                LBTHROW(std::runtime_error("Cannot open BlueConfig file " +
+                                           source));
+            buffer << file.rdbuf();
+        }
+        else
+            buffer << source;
```
hernando: I don't like this implementation very much because it will make errors about inexistent files misleading. What about another constructor that takes a char*?

tribal-tec: I didn't want to advertise this as a real feature, hence hiding it behind the existing API was fine enough for me. It's purpose is not very useful anyways, as the BlueConfig just redirects to other paths. It's nice for a show-case, but in reality you have to open the file close to where all the other referenced entities reside.
So I'm gonna close this then.

hernando: Much more useful would be a remote file browser in Sauron ;)

PR #227, opened 2018-05-25
Sonata reports

hernando:


hernando:
Some changes are preparatory for the upcoming PR about SONATA simulation configurations

brion/plugin/compartmentReportCommon.h
```diff
@@ -33,25 +33,35 @@ class CompartmentReportCommon : public CompartmentReportPlugin
...
+                                 const CompartmentCounts& sourceCounts,
+                                 CompartmentCounts& targetCounts);
 
     virtual bool _loadFrame(size_t frameNumber, float* buffer) const = 0;
     virtual bool _loadFrames(size_t frameNumber, size_t frameCount,
                              float* buffer) const;
 
 private:
-    size_ts _neuronCompartments;
+    mutable size_ts _neuronCompartments;
```
karjonas: Why does this have to be mutable?

brain/detail/circuit.h
```diff
@@ -530,12 +530,22 @@ class SonataCircuit : public Circuit::Impl
...
-        const auto rotationAngleY =
-            nodeGroup.getAttribute<float>("rotation_angle_y", startIdx, endIdx);
-        const auto rotationAngleZ =
-            nodeGroup.getAttribute<float>("rotation_angle_z", startIdx, endIdx);
+        auto getter = [this](const char* attr, size_t start, size_t end) {
+            try
+            {
+                return nodeGroup.getAttribute<float>(attr, start, end);
+            }
+            catch (const HighFive::Exception& e)
```
karjonas: Is there any way to check if the attribute exists instead of throwing and catching exceptions? My issue here is that we might also catch exceptions we actually want to throw.

brion/circuitConfig.cpp
```diff
@@ -64,7 +64,14 @@ struct CircuitConfig::Impl
         : resolver(uri)
     {
         const auto json = parseSonataJson(uri);
-        targetSimulator = json["target_simulator"];
+        try
+        {
+            targetSimulator = json.at("target_simulator");
+        }
+        catch (nlohmann::detail::exception&)
```
karjonas: I think you should catch a more specific exception. I think the out_of_range is the right one since you only want to catch the missing key exception.

brion/circuitConfig.cpp
```diff
@@ -19,208 +19,71 @@
...
-    const brion_nlohmann::json& json, const std::string& networkType,
+    const nlohmann::json& json, const std::string& networkType,
     const std::string& elementName, const std::string& typeName,
-    const boost::filesystem::path& basePath)
+    const PathResolver& resolver)
 {
     std::vector<brion::CircuitConfig::SubnetworkFiles> output;
 
-    const auto nodes = json["networks"][networkType];
+    const auto nodes = json.at("networks").at(networkType);
```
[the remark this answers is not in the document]
hernando: I replaced the previous code with .at because it's already better than crashing when a bad input file is given. Replacing the .at with find + something else wouldn't be better. This function is called from a constructor and you can only make the constructor fail by using an exception, so that something else can only be to throw eventually.
The exception will be propagated out of the brain::Circuit constructor and it's up to the user to catch it or ignore it. I think this is reasonable because the library can't do anything about bad inputs. In the Python wrappings this is enough to not crash the interpreter and having a fairly meaningful error message. I can translate the exception into another type, but I prefer not doing so given that we typically do nothing for these type of fatal errors (other that trying not to crash).

brion/plugin/compartmentReportCommon.h
```diff
@@ -33,25 +33,35 @@ class CompartmentReportCommon : public CompartmentReportPlugin
...
     size_t getFrameCount() const final;
 
 protected:
-    void _cacheNeuronCompartmentCounts(const GIDSet& gids);
+    void _cacheNeuronCompartmentCounts();
     /** @return The frame number of a given timestamp clamped to the simulation
         window. */
     size_t _getFrameNumber(double timestamp) const;
     static GIDSet _computeIntersection(const GIDSet& all, const GIDSet& subset);
+    static size_t _reduceMapping(const GIDSet& sourceGIDs,
```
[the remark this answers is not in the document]
hernando: OK. I thought about it, but I don't see it looking a lot better that way.

brion/plugin/compartmentReportHDF5.h
```diff
@@ -63,37 +65,57 @@ class CompartmentReportHDF5 : public CompartmentReportCommon
...
+
+    void _readMetaData();
+    /** Parses the GIDs and offsets and derives per cell compartment counts.
+        The data from the H5 file is resorted if needed. */
+    void _parseBasicCellInfo();
+    void _processMapping();
+
+    void _writeMetadataAndMapping();
+    void _allocateDataSet();
+    //    void _createMetaData();
```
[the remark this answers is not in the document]
hernando: done

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -577,38 +562,90 @@ void CompartmentReportHDF5::_createMetaData()
...
-    attribute = dataset.createAttribute<double>(dataAttributes[1], scalar);
-    attribute.write(getStartTime());
+    HighFive::DataSpace dataspace(std::vector<size_t>{frames, _frameSize});
 
-    attribute = dataset.createAttribute<double>(dataAttributes[2], scalar);
-    attribute.write(getEndTime());
+    HighFive::DataSetCreateProps chunking;
+    auto chunkDims = _computeChunkDims(_cellSizes, 5);
+    if (chunkDims[0] >= frames)
+        // Only apply chuking if the dataset is large enough. Otherwise HDF5
```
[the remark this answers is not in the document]
hernando: done

brion/plugin/compartmentReportCommon.h
hernando: doesn't need to

brion/circuitConfig.cpp
hernando: Done

brain/detail/circuit.h
hernando: Yes, but I won't do that. The alternative is to get a vector with all object names (datasets or not) and then do a linear search on that list. The resulting code would look less clear.

brion/plugin/compartmentReportCommon.h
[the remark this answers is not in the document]
hernando: I've been trying to do this change properly and it's getting nasty. We can discuss it tomorrow if you want, but minimizing copies without writing strange code requires a lot of changes somewhere else for what's worth it.
I've changed it slightly to get rid of two parameters at least.

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -350,222 +200,357 @@ void CompartmentReportHDF5::writeHeader(const double startTime,
...
-        tmp << datasetName.substr(1);
-        uint32_t gid;
-        tmp >> gid;
-
-        auto dataset = _openDataset(*_file, gid);
-        dataset.getAttribute(dataAttributes[1]).read(_startTime);
-        dataset.getAttribute(dataAttributes[2]).read(_endTime);
-        dataset.getAttribute(dataAttributes[3]).read(_timestep);
+        if (!_verifyFile(*_file))
+            std::string(
```
tribal-tec: throw?

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -350,222 +200,357 @@ void CompartmentReportHDF5::writeHeader(const double startTime,
...
         catch (const HighFive::AttributeException&)
         {
             _tunit = "ms";
         }
+
+        std::vector<double> timeData;
+        time.read(timeData);
+        if (timeData.size() != 3)
+        {
+            throw std::runtime_error(std::string(
```
tribal-tec: rm string

hernando:
Retest this please

hernando:
Retest this please

PR #229, opened 2018-06-07
Morphology recentering for SONATA circuits.

hernando:
Added Circuit::getAttribute to retrieve custom node attributes
from SONATA circuits.


brain/circuit.cpp
```diff
@@ -28,6 +28,277 @@
...
+        CachedMorphologies cached = _loadFromCache(keys);
+
+        struct MorphologyUse
+        {
+            brion::MorphologyPtr original;
+            brion::MorphologyPtr recentered;
+            size_t numRecentered = 0;
+            size_t numOriginal = 0;
+        };
+        using Loading = std::unordered_map<std::string, MorphologyUse>;
```
karjonas: Why use a type alias?

hernando: In the beginning I thought I was going to use this type more than once.

hernando:
I'm trying to provide a test (despite it's quite annoying), but I won't be able to test all combinations. To me this is the type of code that is better to review thoroughly rather than rely on tests and coverage. The tests need to be white box to guarantee coverage and given the amount of combinations 100% coverage doesn't guarantee very much.

hernando:
Done. I also fixed a couple of bugs I found out while doing it and testing with the sample circuit I have

hernando:
Retest this please

tests/brain/sonata_circuit.cpp
```diff
@@ -116,9 +116,47 @@ BOOST_AUTO_TEST_CASE(getMorphologyURIs)
...
+
+    const auto transform = [](const brain::Matrix4f& m,
+                              const brain::Vector3f& v) {
+        const auto t = m * brion::Vector4f(v[0], v[1], v[2], 1);
+        return t.get_sub_vector<3, 0>();
+    };
+    auto transforms = circuit.getTransforms({0, 2});
+    morphologies =
+        circuit.loadMorphologies({0, 2}, brain::Circuit::Coordinates::global);
+    BOOST_CHECK_EQUAL(morphologies[0]->getSoma().getCentroid(),
```
karjonas: Are all transforms identity in the test case?

hernando: No, how do you reach that conclusion?

PR #234, opened 2018-07-04
Fix GCC 8.1 and Clang 6.0 compilation

karjonas:


brion/detail/meshBinary.h
```diff
@@ -40,6 +40,19 @@ T get(const uint8_t* buffer, size_t& pos)
     return *val;
 }
 
+template <typename T>
+std::shared_ptr<std::vector<T>> readBuffer(const uint8_t* const ptr,
```
hernando: According to CI there must be something wrong with this function. I think it's better to use const T* const ptr and the template parameter and do a static cast at call time.

karjonas: Yes, we talked offline and came to the same conclusion :)

karjonas: I added a cast inside the function instead to avoid the extra typing.

PR #235, opened 2018-08-03
Do not use LBTHROW to avoid leaking lunchbox Log instance

tribal-tec:


brion/circuit.cpp
```diff
@@ -51,10 +51,10 @@ class Circuit::Impl
         fs::path path = source;
         const std::string ext = fs::extension(path);
         if (ext != ".mvd" && ext != ".mvd2")
-            LBTHROW(
-                std::runtime_error("Expecting mvd file format for "
-                                   " circuit file " +
-                                   source));
+            throw std::runtime_error(
+                "Expecting mvd file format for "
+                " circuit file " +
```
hernando: merge strings if possible

PR #237, opened 2018-08-16
Do not use LBTHROW to avoid leaking lunchbox Log instance

tribal-tec:


tribal-tec:
You get exceptions, it's up to the caller to print them if needed. In Brayns that's the case.
The log instance is deleted with lunchbox::exit(), but Brion has no semantic like that.

PR #238, opened 2018-08-20
Add support for specifying random seed in method call

karjonas:
I want this so I can specify my own seed when calling this from Brayns without having to set the BRAIN_CIRCUIT_SEED environment variable.

### DIFF
--- a/brain/detail/circuit.h
+++ b/brain/detail/circuit.h
@@ -141,14 +141,14 @@ std::vector<bool> _getAttribute<bool>(const brion::NodeGroup& nodeGroup,
 }
 
 template <typename T>
-void _shuffle(T& container, const std::pair<bool, size_t>& seed)
+void _shuffle(T& container, const size_t* seed)
 {
     std::random_device randomDevice;
     std::mt19937_64 randomEngine(randomDevice());
     const char* seedEnv = getenv("BRAIN_CIRCUIT_SEED");
-    if (seed.first)
+    if (seed)
     {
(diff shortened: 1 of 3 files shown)


brain/detail/circuit.h
```diff
@@ -141,12 +141,16 @@ std::vector<bool> _getAttribute<bool>(const brion::NodeGroup& nodeGroup,
 }
 
 template <typename T>
-void _shuffle(T& container)
+void _shuffle(T& container, const std::pair<bool, size_t>& seed)
 {
     std::random_device randomDevice;
     std::mt19937_64 randomEngine(randomDevice());
     const char* seedEnv = getenv("BRAIN_CIRCUIT_SEED");
```
hernando: Isn't the environment variable enough for your use case?

brain/detail/circuit.h
```diff
@@ -398,7 +402,8 @@ class Circuit::Impl
 
     virtual GIDSet getGIDs(const std::string& target) const = 0;
 
-    GIDSet getRandomGIDs(const float fraction, const std::string& target) const
+    GIDSet getRandomGIDs(const float fraction, const std::string& target,
+                         const std::pair<bool, size_t> seed = {false, 0}) const
```
hernando: Use size_t* size = nullptr for the optional seed parameter, that's a much more common idiom.

karjonas:
I am not sure of other projects but it wont be needed for Brayns anymore.

PR #239, opened 2018-08-22
SONATA compartment report performance improvements

hernando:


brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -637,16 +757,123 @@ void CompartmentReportHDF5::_allocateDataSet()
...
+        _data.reset(
+            new HighFive::DataSet(_file->getDataSet("data", accessProps)));
+        return;
+    }
+
+    // We pick the number of slots so it can at least hold a full trace or
+    // a full frame. Note that the maximum cache size allowed by this number
+    // and the actual maximum cache size are independent.
+    // To avoid collisions reading traces we also need a number of slots s,
+    // such as blocks[1] ∤ s. At the same time we want blocks[0] ∤ s.
```
karjonas: Write "blocks[1] does not divide by s" to avoid using utf-8 symbols.

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -143,7 +187,10 @@ bool CompartmentReportHDF5::handles(const CompartmentReportInitData& initData)
 std::string CompartmentReportHDF5::getDescription()
 {
     return "SONATA HDF5 compartment reports:"
-           "  [file://]/path/to/report.(h5|hdf5)";
+           "  "
+           "[file://]/path/to/"
+           "report.(h5|hdf5)[?[cache_size=(auto|num_bytes):][cell_to_frames="
```
tribal-tec: code uses cell**S**_to_frames

tribal-tec: maybe try to put spaces differently to fix this horrendous clang formatting here?

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -89,6 +98,38 @@ bool _verifyFile(const HighFive::File& file)
...
 
+size_t _parseCacheSizeOption(const URI& uri)
+{
+    const auto i = uri.findQuery("cache_size");
+    if (i == uri.queryEnd())
+        return 0;
+
+    const auto& value = i->second;
+    if (value == "auto")
+        return std::numeric_limits<size_t>::max();
```
tribal-tec: don't know too much about the intrinsics here, but this is more like 'max' cache size. Shall it not be depend on the size of the files or number of rows/frames/cells? 'auto' would imply something 'smart' to me.

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -260,6 +307,48 @@ bool CompartmentReportHDF5::writeFrame(const uint32_t gid, const float* values,
     return true;
 }
 
+bool CompartmentReportHDF5::writeFrame(const GIDSet& gids, const float* values,
+                                       const size_ts& sizes,
+                                       const double timestamp)
+{
+    if (gids.empty())
+        return true;
```
tribal-tec: is more like 'false' to me intuitively

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -308,18 +397,47 @@ bool CompartmentReportHDF5::_loadFrame(const size_t frameNumber,
...
     }
 
-    std::lock_guard<std::mutex> mutex(detail::hdf5Mutex());
-    size_t offset = _sourceMapping.cellOffsets[_subsetIndices[0]];
-    const auto& slice = _data->select({frameNumber, offset},
-                                      {frameCount, _targetMapping.frameSize});
-    slice.read(buffer);
-    return true;
+    // Cheking if the target GIDs are a single slice in the source file.
```
tribal-tec: typo 'Cheking'

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -637,16 +757,123 @@ void CompartmentReportHDF5::_allocateDataSet()
...
     _data.reset(new HighFive::DataSet(
-        _file->createDataSet<float>("data", dataspace, chunking)));
+        _file->createDataSet<float>("data", dataspace, chunking, caching)));
 
     detail::addStringAttribute(*_data, "dunit", _dunit);
 }
+
+void CompartmentReportHDF5::_parseWriteOptions(const URI& uri)
+{
+    for (auto i = uri.queryBegin(); i != uri.queryEnd(); ++i)
```
tribal-tec: const auto&

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -637,16 +757,123 @@ void CompartmentReportHDF5::_allocateDataSet()
...
+    {
+        const auto& key = i->first;
+        const auto& value = i->second;
+        if (key == "cells_to_frames")
+        {
+            if (value == "inf")
```
tribal-tec: it's not documented that 'inf' is valid value

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -637,16 +757,123 @@ void CompartmentReportHDF5::_allocateDataSet()
...
+                }
+            }
+        }
+    }
+}
+
+void CompartmentReportHDF5::_reopenDataSet(size_t cacheSizeHint)
+{
+    // Getting the chunk dims
+    auto properties = H5Dget_create_plist(_data->getId());
```
tribal-tec: HighFive has no support for that?

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -51,6 +51,11 @@ std::vector<hsize_t> _computeChunkDims(const std::vector<uint32_t>& cellSizes,
                                        float cellsToFramesRatio,
                                        size_t blockSize = _blockSize)
 {
+    if (cellsToFramesRatio == 0)
```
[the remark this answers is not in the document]
hernando: I typically use integers and rely on the compiler to do the right thing. But it may be a bad practice on my side.
I've added the check in the parsing code, so this function will never get a negative number. Do you want an assert?

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -89,6 +98,38 @@ bool _verifyFile(const HighFive::File& file)
     return true;
 }
 
+size_t _parseCacheSizeOption(const URI& uri)
+{
+    const auto i = uri.findQuery("cache_size");
```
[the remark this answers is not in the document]
hernando: It's an iterator to a std::pair<std::string, std::string> pair. I could call it keyValueIter.

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -89,6 +98,38 @@ bool _verifyFile(const HighFive::File& file)
...
+    if (pos == value.size() - 1)
+    {
+        if (value[pos] == 'K')
+            size *= 1024;
+        else if (value[pos] == 'M')
+            size *= 1024 * 1024;
+        else
+            size = 0;
+    }
+    if (size == 0 and pos != value.size())
```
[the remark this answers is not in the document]
hernando: No because size == 0 is a valid input, it's used for disabling the cache (which may be beneficial in some situations)

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -116,7 +158,9 @@ CompartmentReportHDF5::CompartmentReportHDF5(
             _parseBasicCellInfo();
             _subset = false;
         }
+        return;
     }
```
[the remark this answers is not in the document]
hernando: I don't have a strong opinion here, but Stefan favored this coding style and it's used thorough all the library. If I do what you request, Daniel you suggest exactly the opposite.

brion/plugin/compartmentReportHDF5.cpp
hernando: Is it a concern having utf-8 symbols in 2018?

brion/plugin/compartmentReportHDF5.cpp
hernando: It got formatted that way because I had a single string. If I split it by hand, clang-format won't break it.

brion/plugin/compartmentReportHDF5.cpp
hernando: Not to me. If you ask me to read nothing, I will succeed reading nothing. Why should it be a failure?

brion/plugin/compartmentReportHDF5.cpp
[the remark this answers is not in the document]
karjonas: It is fine, you can leave it as is.

brion/plugin/compartmentReportHDF5.cpp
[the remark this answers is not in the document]
karjonas: Okay, lets keep it.

brion/plugin/compartmentReportHDF5.cpp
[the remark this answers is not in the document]
karjonas: I think it, iter or keyValueIter is fine.

brion/plugin/compartmentReportHDF5.cpp
hernando: It's something smart, but I found a bit annoying to document it in the description. Auto means allocating space for a whole frame or trace, whichever is bigger. It's not really max, std::numeric_limits<size_t>::max() is used as a magic value to identify this request in the client code. To make it more semantic I've replaced it with a constant.

brion/plugin/compartmentReportHDF5.cpp
hernando: In this case it doesn't matter, but since this is just an iterator, don't you think that adding a reference would make the code slower for a long loop? Even with optimizations I have the feeling that the code would have more pointer chasing with the reference.

brion/plugin/compartmentReportHDF5.cpp
hernando: No, because it's a getter and we have started adding support for different properties quite recently. For the few number of lines it is I prefer not messing with HighFive.

brion/plugin/compartmentReportHDF5.cpp
hernando: Done, I've also documented what cache_size=auto does.

hernando:
Any more comments on this one?

karjonas:
lgtm

PR #244, opened 2018-09-14
Implemented SONATA section ID assigment in swc morphology parsing

hernando:


brion/plugin/morphologySWC.cpp
```diff
@@ -80,15 +76,24 @@ struct Sample
...
+    int nextID = -1;        // sample index
+    int siblingID = -1;     // sample index
+    int sectionID = -1;     // Only necessary for the last sample of a section
+    int parentSection = -1; // Only meaningful for the first sample of a section
+};
+using Samples = std::vector<Sample>;
+auto _comparator = [](const Sample* a, const Sample* b) {
+    // We want undefined sections to go last but we need to respect the
+    // numerical values from the swc "specification".
+    // The conversion of -1 to size_t takes care of this
```
mgeplf: uint32_t?

Neat trick, but I wonder if would be better to be explicit about handling SWCSectionType::SWC_SECTION_UNDEFINED, that way it's safe against refactors and such.

tests/data/swc/section_order.swc
```diff
@@ -0,0 +1,37 @@
+1 1 0 0 0 10 -1
+
+# section 11
+2 4 0 0 1 0.5 -1
```
mgeplf: 1?

hernando: You mean the section ID should be 1. Nope, this section is an apical dendrite so it goes last.

brion/plugin/morphologySWC.cpp
```diff
@@ -80,15 +76,24 @@ struct Sample
...
+    return (uint32_t(a->type - 1) > uint32_t(b->type - 1) ||
+            (a->type == b->type && a > b));
```
wizmer: Why do you use pointer comparison in case of equality ?

tests/data/swc/section_order.swc
mgeplf: Sorry, I forgot that it's Gerrit that highlights what I was looking at.  I mean the parent ID should 1, since I assume it's connected to the soma

brion/plugin/compartmentReportHDF5.cpp
```diff
@@ -522,8 +522,15 @@ void CompartmentReportHDF5::_parseBasicCellInfo()
     catch (...)
     {
     }
-    if (*offsets.rbegin() >= _sourceMapping.frameSize)
+    if ((offsets.size() == gids.size() &&
+         *offsets.rbegin() >= _sourceMapping.frameSize) ||
+        // Allowing offsets to be have one more position to swallow
+        // some non-conforming files.
```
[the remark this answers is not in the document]
hernando: The only instance is a file created by someone from the Allen Institute. I can remove this change and add it back if really needed. I must admit that I was sneaking it to not have to worry about it in the future.

brion/plugin/morphologySWC.cpp
```diff
@@ -80,15 +76,24 @@ struct Sample
...
-    int parentSection; // Only meaningful for the first sample of each
-                       // section
+    int parent = -1;        // sample index
+    int nextID = -1;        // sample index
+    int siblingID = -1;     // sample index
+    int sectionID = -1;     // Only necessary for the last sample of a section
+    int parentSection = -1; // Only meaningful for the first sample of a section
+};
+using Samples = std::vector<Sample>;
+auto _comparator = [](const Sample* a, const Sample* b) {
```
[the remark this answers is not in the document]
hernando: That's not so easy, std::priority_queue doesn't accept functions in the template instantiation, it has to be a class with a operator<. Do you really want to have that?

brion/plugin/morphologySWC.cpp
hernando: The alternative would be to add if's to the code and they make me sad.
I think the current code is safe against refactorings because, as the comment states, the numeric values of the enum values come from the SWC spec and we can't change that. Otherwise I would have moved UNDEFINED to another position in the enum declaration and done.

brion/plugin/morphologySWC.cpp
hernando: Good question, that should be documented. The address reflects which sample was parsed first from the file, which is exactly the sorting condition that we have to apply to follow the spec.

tests/data/swc/section_order.swc
hernando: Yes, and no. The implementation is resilient to first order branches not being connected to anything. I didn't connect this one to the soma on purpose.

(Despite all harsh criticism that gerrit gets for it's UI, I still think that code reviews are more effective there)

brion/plugin/morphologySWC.cpp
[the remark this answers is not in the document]
mgeplf: IIRC, it does accept functions, can't you do: `bool _comparator(const Sample* a, const Sample* b) {....}`

Then: `using SampleQueue = std::priority_queue<int, std::vector<int>, decltype(&_compare)>;`

However, I guess that makes the instantiation of the queue a bit more complicated: `SampleQueue q(_compare)`

Anyways, I guess your style guide is cool w/ this `auto` usage, so then it's cool w/ me

tests/data/swc/section_order.swc
mgeplf: Ok, fair enough.  Maybe leave a comment so the reasoning is captured.

Agree about the gerrit review

brion/plugin/morphologySWC.cpp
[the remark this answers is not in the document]
hernando: OK, I see, I was missing the & in decltype(&_comparator)

hernando:
@rdumusc, can you approve? I think the others can't do it.

PR #250, opened 2018-11-15
Random gids

hernando:


brain/detail/util.h
```diff
@@ -0,0 +1,66 @@
...
+namespace brain
+{
+namespace
+{
+template <typename T>
+inline void _shuffle(T& container, const size_t* const seed)
+{
+    std::random_device randomDevice;
+    std::mt19937_64 randomEngine(randomDevice());
+    const char* seedEnv = getenv("BRAIN_CIRCUIT_SEED");
```
karjonas: Can we remove this special environment variable?

hernando: I didn't add it. Any comment, @tribal-tec?

tribal-tec: If it's so annoying to people, it can be removed. Commandline it is then I guess? Documentation/scripts which use the env variable shall be adapted accordingly then.

hernando: I doesn't annoy me at all. If there's anything depending on it, I wouldn't remove it.

PR #263, opened 2019-07-18
Detect start.target in CircuitPath

### LINKED ISSUE #261: Detect `start.target` in CircuitPath
Neurodamus supports `nrnPath` in BlueConfig pointing to a file rather than to a directory now.
`start.target` auto-detection should take that into account and be able to either:
 * always look for `start.target` at `CircuitPath` rather than `nrnPath` (would work for all circuits we are working with)
 * look for `start.target` at `CircuitPath` if it can not find it at `nrnPath`
 * look for `start.target` at `dirname(nrnPath)` if `nrnPath` points to a file

karjonas:
Fixes #261

### DIFF
--- a/brion/blueConfig.cpp
+++ b/brion/blueConfig.cpp
@@ -73,7 +73,7 @@ inline std::string lexical_cast(const brion::BlueConfigSection& b)
     }
     throw boost::bad_lexical_cast();
 }
-}
+} // namespace boost
 
 namespace brion
 {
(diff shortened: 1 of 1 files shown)


brion/blueConfig.cpp
```diff
@@ -213,16 +213,14 @@ class BlueConfig
...
 
 BlueConfig::BlueConfig(const std::string& source)
     : _impl(new detail::BlueConfig(source))
 {
 }
 
-BlueConfig::~BlueConfig()
-{
-}
+BlueConfig::~BlueConfig() = default;
```
mgeplf: Does it make sense to put this in the header? (I don't know what style you guys use)

karjonas: Maybe, I am not sure if there is a reason for having it here or in the header, so I will keep it for this PR.

PR #275, opened 2020-04-30
Stripping Keyv lib from Brion

NadirRoGue:


ppodhajski:
perhaps a line in changelog would be useful as we remove this functionality?

NadirRoGue:
> perhaps a line in changelog would be useful as we remove this functionality?

Done. Also added the major changes since last changelog update

PR #278, opened 2020-05-07
Stripping vmmlib from Brion. Replaced with GLM

NadirRoGue:


ppodhajski:
retest this please

NadirRoGue:
retest this please

brain/circuit.cpp
```diff
@@ -49,17 +50,18 @@ URIs _getMorphologyURIs(const Circuit::Impl& circuit, const GIDSet& gids)
...
         throw std::runtime_error(
             "Positions not equal rotations for given GIDs");
 
     Matrix4fs transforms(positions.size());
 
-#pragma omp parallel for
+    #pragma omp parallel for
     for (size_t i = 0; i < positions.size(); ++i)
-        transforms[i] = Matrix4f(rotations[i], positions[i]);
+        transforms[i] = glm::translate(glm::mat4(1.f), positions[i]) * glm::mat4_cast(rotations[i]);
```
chevtche, line 63: Are you sure ? This shit can be tricky....

brain/circuit.cpp
```diff
@@ -275,13 +277,13 @@ class MorphologyLoader
         const size_t end = somaID == sections.size() - 1
                                ? points.size()
                                : sections[somaID + 1][0];
-        Vector3f centroid;
+        glm::vec3 centroid;
         for (size_t i = start; i != end; ++i)
-            centroid += points[i].get_sub_vector<3, 0>();
+            centroid += glm::vec3(points[i]);
```
chevtche, line 282: Not sure this is right.

brain/circuit.cpp
NadirRoGue: I tested both ways. The rotation then translate was wrong. This one yields the same result as with vmmlib

brain/circuit.cpp
NadirRoGue: The get_sub_vector<Length, Start> returns a vector of Length whose coefficents are taken from the buffer starting at Start. So it takes the 3 first components of the 4 component vector. The constructor for vec3 does the same.

brain/detail/circuit.h
```diff
@@ -96,15 +96,16 @@ void _assign(const ::MVD3::Range& range, const GIDSet& gids, SrcArray& src,
...
-    return Vector3f(subarray[0], subarray[1], subarray[2]);
+    return glm::vec3(subarray[0], subarray[1], subarray[2]);
 }
 
-Quaternionf _toQuaternion(
+glm::quat _toQuaternion(
     const ::MVD3::Rotations::const_subarray<1>::type& subarray)
 {
-    return Quaternionf(subarray[0], subarray[1], subarray[2], subarray[3]);
+    return glm::angleAxis(static_cast<float>(subarray[0]),
```
chevtche, line 107: Test it

PR #281, opened 2020-05-20
Stripped CMake/common (with Pydoxygen and Servus)

NadirRoGue:
Stripped CMake/common from Brion
Pydoxygen and Servus could be easly replaced as well, and heavily depended on CMake/common, so they have been stripped as well


ppodhajski:
removed eyescale dependencies and patch from spack package.py.

ppodhajski:
retest this please

NadirRoGue:
retest this please

PR #284, opened 2020-06-02
Fix installation structure from CMake

NadirRoGue:


deps/CMakeLists.txt
```diff
@@ -14,10 +14,32 @@ if(NOT glm_FOUND AND NOT TARGET glm)
     set(CMAKE_EXPORT_NO_PACKAGE_REGISTRY ON)
     set(CMAKE_INSTALL_LIBDIR lib)
     add_subdirectory(glm)
+
     # WAR for https://github.com/g-truc/glm/issues/854
     if(CMAKE_COMPILER_IS_GNUCXX AND NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 8.0)
       target_compile_options(glm INTERFACE -Wno-error=class-memaccess)
     endif()
+
+    #add_library(glm INTERFACE)
```
ppodhajski, line 23: I think commented out parts should be removed

NadirRoGue: Done

PR #332, opened 2021-11-15
Improving submodule handling. Fix morphio aliasing

NadirRoGue:


deps/CMakeLists.txt
```diff
@@ -1,97 +1,97 @@
-include(ExternalProject)
-
 # Mandatory dependencis not found in the system will be built
```
jplanasc, line 1:
```suggestion
# Mandatory dependencies not found in the system will be built
```

jplanasc (approved):
Looks good to me.

deps/CMakeLists.txt
NadirRoGue: Good catch, thanks!

PR #335, opened 2021-12-13
Fix range-loop-construct warnings.

matz-e:
Complaints about implicit copying that can be avoided with references.
The current code will fail to compile with gcc@11: due to -Werror.


NadirRoGue:
Removed ubuntu16 from CI

retest this please

NadirRoGue (approved):

PR #345, opened 2022-04-12
Fix currentDir handling

mgeplf:
* if CurrentDir isn't in the BlueConfig, it should default to the path
  where the BlueConfig is, using the absolute path in this patch
* if a BlueConfig in the current directory is passed, `adjust_path`
  would be called with  ("BlueConfig", "", ".") if OutputPath  is "."
  this would return the `fullPath = currentDir + "/" + cleanPath; `
  `/.` since currentDir is empty, and this path exists.  This would mean
  the URI wasn't correct

### DIFF
--- a/brion/blueConfig.cpp
+++ b/brion/blueConfig.cpp
@@ -255,9 +255,15 @@ public:
         return std::string();
     }
 
-    const std::string& getCurrentDir()
+    const std::string getCurrentDir()
     {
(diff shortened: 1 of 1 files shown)


NadirRoGue:
Thanks a lot, it has always been something that wasn't fully complying with the doc.

NadirRoGue (approved):

mgeplf:
No worries, thanks for accepting it.

